PR #304, opened 2014-04-18
Bugfix: Potential buffer overflow in memcpy().

rolandbosa:
While scanning the code I came across this particular memcpy(), which seems to use
the wrong function to calculate the number of bytes to copy. The correct function to use
is qMin(), to avoid the buffer overflow.

Also, according to the comments in the header file for the structure, it is ok to fill the field
up to the very last byte and omit the terminating zero-byte. I didn't take the time to go into
the details on the receiving side and validate that the comment actually is implemented.
So to err on the safe side, this code still hardcodes a terminating zero-byte at the end of
the name, just to make sure.

Lastly, this code compiles, but I haven't found a way to exercise it. Please review thoroughly.

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -2534,12 +2534,22 @@ void UAS::requestParameter(int component, const QString& parameter)
     mavlink_param_request_read_t read;
     read.param_index = -1;
     // Copy full param name or maximum max field size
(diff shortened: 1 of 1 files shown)


Arne-W:
The code this fix is for does not exist anymore - nevertheless thank you for your effort.

PR #305, opened 2014-04-18
Experimental: Qt5 / Code reduction and streamlined compass calibration

rolandbosa:
Although I rebased this one on top of master, it should probably go into the Qt5-Wip branch, if at all. It's big and kind of unwieldily, but it packs a lot of punch: it represents a few things that I wanted to fix on the Qt5 side:

* Improve compile times
* Reduce warnings
* Streamline the calibration process to not use message boxes.

In particular, this boiled down to the following...

There's a gazillion of warnings coming from the Eigen and AlgLib libraries. Those libraries also use templates very heavily and are in my opinion one of the reasons why the compile times are so long. So I tried to find out exactly which parts needed those "light-weight" (in my opinion, excessive use of templates is heavy-weight) libraries. It turns out that the Eigen library boils down to a single quaternion rotation to translate a frame of reference into another. The AlgLib library is used only for the compass calibration to perform a least squares reduction of the calibration data. So I refactored out the vector/quaternion stuff and added the least squares reduction into that mix too. It still needs the least squares stuff, but at least now it is in a re-usable Vertex3D API. The Eigen library however has left the building. My benchmarks showed a 8% speed increase in total compilation time. I'm sure there's more to found, but this is a start.

Since I was at the compass calibration page already, I figured I'd try to get rid of the modal progress box and the confirmation box. In my opinion, they are distracting and don't help with the flow of a calibration. So I changed the calibration to have those fields (a progress bar and some status) right there on the field. I also re-arranged some of the controls, fixed the hyperlink to actually open the magnetic declination page and generally aligned and stacked and resized things, so that they look good in a variety of screen resolutions. If the resolution is too small, it now even pops up a nice scrollbar.

So, long story short: try it out and give me some feedback - I don't expect this to go straight into production though.


billbonney:
These changes inspired some recent refactoring if the compass code, thx. Bill :+1:

PR #417, opened 2014-07-24
Comms: Fix to undo tautological code

hughescr:
Re-submitting with static_cast instead of C-style cast


billbonney:
merged with 3d2da0d

PR #504, opened 2014-10-24
QtGStreamer support for PFD (QML Version), and HUD, Experimental feature (updated and refactored)

pnduffy:
This fork includes experimental code for QtGStreamer which incorporates video into the PDF and HUD displays. The new dependencies are GStreamer, QtGStreamer, and Boost. The proposed change should include a new #define that excludes these changes if GStreamer and QtGStreamer are not installed. This will required a change to the make files (not done here). This feature is especially interesting to FPV flyers and enables both normal analog video and HD video to be overlayed with flight data.

To use the feature, a binary install of GStreamer is needed on the local machine. These are available from the GStreamer.freedesktop.org website.  In APM Planner, the 'Settings' must be set to enable GStreamer and a valid GStreamer pipeline string set in the configuration. These settings are in the 'APM Planner' section of the configuration.


billbonney:
Cleanup of old PRs

PR #564, opened 2014-12-27
Reduced warnings emitted from Eigen

Arne-W:
Hi all,
Compiling the APM-Planner on Linux (ubuntu 14.10) shows round about 2043 warnings. I dont 
think its really critical but we should do something about that cause nobody will pay any attention to warnings anymore nor will even recognise when producing them. 
In my opinion warning free compilation leads to better code quality and so should be a goal of development.
So I tried to do something to reduce warnings. The Eigen lib is the worst part as the programmer of the lib knows by himself he created an special include to supress warnings while his library is included. This special include does not provide the warning supression for the GCC. So I extended the precompiler code to support GCC too reducing the warnings to 1007 on my ubuntu.
Yes I know its not a good way to reduce warnings by supressing them, but it is an external lib and not maintained by the apm_planner people so the changes made to that code should be as small as possible.
Sure the way rolandbosa invented is the better one, he removed the eigen lib code completely, but unfortunately nobody was able to check his code for correctness.

Okay some details about the changes measured on Ubuntu 14.10, QTcreator 3.1.1, gcc 4.9.1
Changes of precompiler code -> reduces warnings from 2043 to 1007
removing the RealScalar type -> reduces warnings from 1007 to 880

greetings from Hamburg

### DIFF
--- a/libs/eigen/Eigen/src/Core/util/DisableStupidWarnings.h
+++ b/libs/eigen/Eigen/src/Core/util/DisableStupidWarnings.h
@@ -37,6 +37,13 @@
     #pragma clang diagnostic push
   #endif
   #pragma clang diagnostic ignored "-Wconstant-logical-operand"
+#elif defined(__GNUC__) || defined(__GNUG__)
+  #ifndef EIGEN_PERMANENTLY_DISABLE_STUPID_WARNINGS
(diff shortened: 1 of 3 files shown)


billbonney:
Eigen dependency has been removed, closed

PR #575, opened 2015-01-07
Build: Added Debian compatibility on dependencies.

jaqm:
Ubuntu and Debian use a different name for some packages.
Ubuntu - Debian
1.- libqt5webkit5-qmlwebkitplugin - qml-module-qtwebkit
2.- qtdeclarative5-qtquick2-plugin - qml-module-qtquick2
The result is a package that can be built and installed in Debian and Ubuntu.


billbonney:
merged with 8eb0dd6e055bee1fd4f06a6f30199cb64507acdc

PR #576, opened 2015-01-07
Comm: TCPLink doesn't update UI upon disconnection

nbrachet:
Disconnecting a TCPLink leaves the UI in the "connected" state, and there is no way to reconnect.


billbonney:
merged with 5784673b34897d6fcaf2c571012e860f8ab6e237 Thx :+1:

PR #581, opened 2015-01-14
Plan View: tooltip no longer adds MSL to any altitude.

AndKe:
The WP tooltip said that any altitude were an MSL altitude, regardless
of the wp being relative to home_alt.
This could be dangerous if certain plan types were verified based on tooltip.
I tried to add conditional string, but did not find how to get altitude-type.

### DIFF
--- a/libs/opmapcontrol/src/mapwidget/waypointitem.cpp
+++ b/libs/opmapcontrol/src/mapwidget/waypointitem.cpp
@@ -330,7 +330,7 @@ namespace mapcontrol
     void WayPointItem::RefreshToolTip()
     {
         QString coord_str = QString::number(coord.Lat(), 'f', 6) + "   " + QString::number(coord.Lng(), 'f', 6);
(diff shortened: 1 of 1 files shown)


billbonney:
I'll take a look at this, you can determine if the waypoint is MAV_GLOBAL or MAV_FRAME_GLOBAL_RELATIVE_ALT for MSL or AGL reporting

AndKe:
I admit failed to find out how to get these data, this application seems very object-oriented, and I have problems finding out where/how to properly interact with the data.
Also, few days ago I was missing a total mission length display, and "distance between waypoints" in tooltip - and could not even decide where a function that calculate distance between two coordinates should be put to be available from "anywhere" in the code.    (the idea was to calculate each wp's distance to previous WP when moved.)

billbonney:
I just had a quick look at the WaypointItem and it doesn't encapsulate  the waypoint type. I would add a type parameter and it can be set when the WaypointItem is created, so it can be rendered on the map.

The strange thing is that there are essentially two waypoint lists in the app. ie. the Waypoint list that is associated with the UAS object, managed through he UAS Waypoint Manager. Then the WaypointList of graphical item that is displayed in the MapWidget. Its not a very clean design, but it allowed reuse on a map widget written in a different project

AndKe:
while you are at adding, please add "distance_from_last_wp" 
That would be used to get a total length, and easy planning when a certain distance between grid-lines are needed.  We would also easily see limits between VLOS and EVLOS missions etc.

billbonney:
fixed with 249bc8e

PR #588, opened 2015-01-19
Spec-file and scripts to build RPM

froller:
scripts/prepare_rpm.pl - pulls latest updates from repo, creates source tarball and updates spec-file
scripts/build_rpm.sh - builds RPM package
redhat/apmplanner2.spec - spec-file template


froller:
There's issue in this changes

PR #590, opened 2015-01-21
Spec-file and scripts to build RPM

froller:
scripts/prepare_rpm.pl - pulls latest updates from repo, creates source tarball and updates spec-file
scripts/build_rpm.sh - builds RPM package
redhat/apmplanner2.spec - spec-file template


malcom2073:
Thanks for working on this, I'll give it a shot. What operating systems are you using, Fedora?

froller:
Yep. Fedora 20.
I'm going to test it on f21 as soon as I install one.

malcom2073:
When I try to run it in FC21, prepare succeeds, but build does this:

https://gist.github.com/malcom2073/81816d3dfdd03edac239

Looks like somehow the tar file has clipped the first two letters of each filename.

froller:
Yes, tar modifies filenames to make same tarballs from any directory you have cloned apm_planner

Issue
pwd > out.txt && tar -ztf ~/rpmbuild/SOURCES/apmplanner2-2.0.16.tar.gz >> out.txt
and send me an out.txt please.

Seems it's odd combination of cwd and regex leads to tarball path corruption.

malcom2073:
Done: https://gist.github.com/malcom2073/5f310ee050042d9daa7e

froller:
Dots in relative path name were not escaped. Thus they're treated as regex macros and ".." matched any two characters instead of ".." only.

froller:
Malcolm, could you test build scripts on f21?

malcom2073:
Yep, seems to work on Fedora 21. If you want to rebase your changes on top of the latest master, I'll get it pulled in (merges are bad).

PR #604, opened 2015-02-06
Some waypoints are not drawn with path. And ROI is visible on Map Widget BugFix.

wolkstein:
This pull request allows waypoints in a QList insert. For these waypoints listed here then a path is drawn.

Before:
differentloiterandland-orig

After:
differentloiterandland-fork

### DIFF
--- a/src/Waypoint.cc
+++ b/src/Waypoint.cc
@@ -75,12 +75,23 @@ Waypoint::Waypoint(const Waypoint& waypoint)
 }
 
 Waypoint::~Waypoint()
-{    
+{
 }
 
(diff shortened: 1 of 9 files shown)


wolkstein:
in the second part of pull request i fix the roi bug. Set Roi Waypoints now visible on map. i split this into three commits. 819db49 a0002cf 5840957

hopefully this is ok for you:)

/g
wolke

billbonney:
Thanks, I squashed the commit and is fixed in biils-wip, which will go into master today. :+1: 
see ac4c3cf

PR #622, opened 2015-02-15
Linux: removed extraneous paths and QT4 reference

stephendade:
Removed extraneous build path reference. Also removed QT4 from autobuild
script.
Thanks to @gmorph for spotting it


billbonney:
merged with 3f7633f

PR #632, opened 2015-03-05
Audio: use "default" ALSA device instead of "plughw:0"

xmxwx:
There is no reason to enforce playback via hardware device instead of the
default ALSA device configured in the system. Such a default device could be a
PulseAudio plugin for ALSA which allows ALSA clients to transparently use
PulseAudio - benefiting from its software mixing and preventing "Device or
resource busy" error when openning the audio device.


billbonney:
Thx for the fix :+1:

PR #633, opened 2015-03-05
Master qt4

kbalisciano:
Please ignore the readme since I made it specific to 12.04 Ubuntu with QT4 with qreal compiled as double on an ARM board. 

I had to make changes to the Freenect files in order to compile with freenect support. 
#include libfreenect/libfreenect.h wouldn't find the file but 
#include libfreenect.h worked. 

Also, Freenect.cc depends on QGC object which is not included. I don't believe this change is specific to ARM boards but it is worth noting.


billbonney:
master_qt4 branch is just a marker of the last qt4 based build. It no longer in development. All platforms support Qt5.

This pull request is against new master that is QT5. I'm closing this PR unless it made against current master.

PR #651, opened 2015-03-19
Add scrollbar to Unmanned Systems list when list is long

labtoast:
When connecting multiple UAVs the Unmanned Systems list will flow off the widget making some UAVs unreachable. This change puts the list in a scrollable widget so that when the list grows long enough a scrollbar will appear and allow you to reach the lowest UAVs.

Without:
screen shot 2015-03-19 at 6 31 27 pm

With:
screen shot 2015-03-19 at 6 42 35 pm


billbonney:
cherry-picked as af29fe8 Thx :+1:

PR #684, opened 2015-04-23
Removing trailing zeros from mouseover tooltip

AndKe:


AndKe:
sorry, I do not know how/where to implement that,nor I am sure how it works by reading the comment :)

billbonney:
you do something like
````
QString str = QString().sprintf( "%.4g", graph->data()->value(key).value );
````
str being formatted as you would expect

AndKe:
Thanks, I am submitting a new pull request (just because I don't have time to figure out if it's possible to "add" to the first version of the fix.)

AndKe:
please disregard this pull request, and accept https://github.com/diydrones/apm_planner/pull/687 instead.
(In the mean time, I've learned that I could have fixed this one instead of creating a new.)

PR #687, opened 2015-04-25
Graph: Removed trailing zeros from mouseover tooltip

AndKe:


billbonney:
I used only 8c5b010 as the code is easier to read, and not less efficient.

PR #692, opened 2015-04-26
Airspeed setup: adding Pixhawk I2C airspeed sensor

AndKe:
User had choice between EagleTree or MEAS I2C Airspeed sensors, even
both use same value for config, there were no intiutive way to choose
the most common 3DR's Pixhawk Airspeed I2C sensor.


billbonney:
I cam merge this in, but the MEAS I2C is the 3DR Airspeed Sensor. Maybe it's juts needs rewording?

AndKe:
I considered changing the shortest (MEAS) to "MEAS/Pixhawk I2C")  - but what would be longer than the dropbox, possibly causing a place problem on small screens.
Another solution would be to have only "Analog" or "Any I2C"  which would do the job just fine now, but it's not future-proof.  
So I figured the teaspoon method in PR would be a good choice.
I can change it to "MEAS/Pixhawk I2C" if you don't think that'll be too long and/or mess up symmetry of the width of those three droplists.

billbonney:
closed with 43d5f9d

PR #694, opened 2015-04-27
Mission Widget: button hotkeys fix

AndKe:
By some mistake, Ctrl-L was assigned to wrong button.


billbonney:
I know I have added ctrl+w as write in the other request, but we should think about that, as ctrl-w is 'close window' as standard for Qt based apps.

Any suggestions to another key instead if ctrl-w

Please feel free to ping me in gitter 

[Gitter](https://gitter.im/diydrones/apm_planner?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)

PR #696, opened 2015-04-28
Plan View: Move mission item to Top/Bottom

AndKe:
Simply move an item to top/end of flightplan.


billbonney:
Closed with 0fd6525

It might be cool to not use the extra button in the view, but have the button option changes if you hold down CTRL for example?

AndKe:
Yes, that would be another solution, but I were not sure is changing QPushbutton.Icon on Ctrl press would apply the change instantly (to make the option clearly available without forcing the user to actually test shift/ctrl +click to guess).   Also I figured we had plenty of horizontal space left, and could as well show it clearly to the user.   
Finally - I am no good at GUI related stuff, and much of the magic used to even make new buttons work is a mystery to me :)
Thanks for accepting the change.

Issue #699, opened 2015-05-01
Graph: time_boot_ms and time_usec fields from tlog files are not showing in the grid view


nbrachet:
Steps to reproduce:
1- Open a TLog file
2- Any MAVLink message with a time_boot_ms or time_usec as the first field (SCALED_IMU2 or GPS_RAW_INT for instance) is not showing a value in the grid

APMPlanner 2.0.17 (249bc8e)

PR #700, opened 2015-05-01
Quick&Dirty fix for #699

nbrachet:
Make sure to include the first field in the returned list, even if it is used to track time


billbonney:
closed with 4cad40f

billbonney:
This fixed caused an issue of 64Bit build on linux as the compiler complained that uint64_t conversion was ambiguous. unsigned long int != unsigned long long ie uint64_t != quint64 on linux 

OSX worked ok.

PR #736, opened 2015-06-14
Joystick improvements

### LINKED ISSUE #718: Joystick Support: Add 'save settings' feature
https://github.com/diydrones/apm_planner/issues/144 requires addition of code to persist settings on reboot. The settings should be saved based on Joystick Identifier.

mrpilot2:
Fixed #718 Add 'save settings' feature - joystick settings are now saved based on joystick id

Improvements:
- create joystick button labels dynamically according to number of buttons of the current joystick
- change label color only if the button state changes to avoid flickering when the color got reset with a timer


billbonney:
Thx, merged. But i still have a comment about the dynamism of the joystick detection. I don't usually have much time to play with joystick support, so efforts here are appreciated :+1:

mrpilot2:
Further improvement I'm thinking about:
Currently  the init method of the joystick blocks until a joystick is detected. After this it is not possible to detect if a joystick gets connected/disconnected. A 'joystick detector' should be implemented that periodically checks if a joystick gets connected/disconnected and show a highly visible warning in case the joystick gets disconnected and it's activated to control the UAS.

I'm currently in the phase of building my first APM drone so at this point of time I can only work on GCS issues because I can't test with real hardware.

PR #738, opened 2015-06-15
Joystick improvements

mrpilot2:
Hi,

as promised, here is my bugfix for the joystick initialization issue

I've put the init method back to the run method (it was there before my changes), which blocks until a new joystick is detected. In the init method whenever a new joystick is detected the settings are loaded and the mappings and buttons on the UI are updated.

I've found out that disconnecting and connecting the joystick again does not work (because the joystick init method is just called once) which didn't work either before my changes (commit bca8b5c085). There is some improvement needed. I'll create an issue for that.


billbonney:
I pulled in the change and pushed to master. I'm closing this PR, rasie a new one when you have the new chnages. :+1:

PR #759, opened 2015-07-03
Elevation Widget: added distance from previous WP.

AndKe:
This adds distance from previous waypoint. Very useful when user needs
to verify/plan a survey grid for proper overlap.

### DIFF
--- a/src/ui/MissionElevationDisplay.cpp
+++ b/src/ui/MissionElevationDisplay.cpp
@@ -272,7 +272,7 @@ void MissionElevationDisplay::addWaypointLabels()
         customPlot->addItem(itemTracer);
 
         QCPItemText *itemText = new QCPItemText(customPlot);
-        itemText->setText(QString::number(wp->getId()));
+        itemText->setText("WP" + (QString::number(wp->getId())) + " (+" + (QString::number(distance,'f', 1)) +"m)");
         itemText->position->setParentAnchor(itemTracer->position);
 
         double adjustedAlt = 0.0;


billbonney:
I did some work to add markers as well, though for some reason I get a lock-up... so i didn't publish
elevation

billbonney:
PS: you can check out the whip in this branch https://github.com/diydrones/apm_planner/tree/elevationmarker-wip

AndKe:
Yes, I had a line-break initially too but hiding "WP" may cause confusion.  Like your "WP2" is outside viewable area in the screenshot above.
As i wrote in the chat - I have not had success increasing the X,Y axis by modifying the "range" operations.

billbonney:
setRange is called 4 times, the first two calls are when the widget is first created. The second two are called when a new WP is added, one to extend X and the other Y. And in anywise, this doesn't cause a redraw until replete is called. The issue you might have add is the call to rescaleAxes() which negates yAxis changes I suspect.

billbonney:
And secondly, if you look at this line of code in the elevationmarker-wip https://github.com/diydrones/apm_planner/commit/90647f2e7086ba7c552c8783650959f554130c52#diff-44b14e52a715728ce54ce40b5012c936R296 
you will see that the code creates two text items, one above the point as WP0 and one below as the distance. The text is anchored to the point, it just needs to be anchored via the top of the text, not the bottom. 
The possible fixes are 
1) add a buffer to the height of the graph, similar to how the xAxis works
2) make the highest point show the WP test below the point and the distance item anchored below that
3) make all WP text points below the line, which is OK until a point is actually on the xAxis line.

I'll look at it some more to make it look better, even just changing the lines to be dashed grey lines would help it clutter the screen less.
Thanks for the input your changes spurred me on to make it better :+1:

AndKe:
What I did try is to change all *Axis->setRange to (first percentage increase)  hen just foolproof huge increase , like from  
    customPlot->yAxis->setRange(ElevationDefaultAltMin,ElevationDefaultAltMax); //m
to 
    customPlot->yAxis->setRange(ElevationDefaultAltMin,(ElevationDefaultAltMax+5000)); //m

I got no different result. - so I guessed "QCPRange xRange = customplot->xAxis->range();" had something to do with that.

PR #760, opened 2015-07-07
Map Widget: Fixes #701 Add auto map caching option

### LINKED ISSUE #701: Maps caching
While caching a larger area for offline usage APM Planner caches several zoom levels, which might take a long time....

While the time it takes is not "controlled" by APM Planner, one setting on APM Planner is kind of annoying, and could be changed.

Currently, after a zoom level is cached, APM Planner stops and asks if the following zoom level is to be cached and just sits there.....This dialog box could be eliminated of have a timeout with a default of yes (keep caching). MP on the other hand just goes to all zoom levels without asking.

It's irrelevant from which map provider this is done. On the pic I'm doing from Bing, because the coverage is newer than GGL :)

screen shot 2015-05-02 at 03 07 21

sutherlandm:
Add option when caching maps to auto continue to a desired zoom level. Form allows for both single level and multi-level caching and prompts user at the end of either caching session to stop or continue. Desired zoom level can be changed at any time (unless user prompt is displayed) and will stop at or continue to new desired level. 

map1

map2

map3

### DIFF
--- a/libs/opmapcontrol/src/mapwidget/mapripform.cpp
+++ b/libs/opmapcontrol/src/mapwidget/mapripform.cpp
@@ -33,21 +33,68 @@ MapRipForm::MapRipForm(QWidget *parent) :
     ui(new Ui::MapRipForm)
 {
     ui->setupUi(this);
+
+    //Hide the progress bars and resize until ripping time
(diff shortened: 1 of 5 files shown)


billbonney:
Thanks for the PR :+1:

PR #762, opened 2015-07-09
Updated the firmware uploading information to include the Gumstix AeroCore

smithandrewc:
Add support for the Gumstix AeroCore.


billbonney:
added with commit 7b34a00 Thx :+1:

PR #763, opened 2015-07-09
Radio Calibration: lower minimum PWM value

lucasdemarchi:
Trying to calibrate a Turnigy 6XS radio is not possible because we get
values around 870usec on some channels. Lower the minimum values so we
can calibrate them.

### DIFF
--- a/src/ui/configuration/RadioCalibrationConfig.h
+++ b/src/ui/configuration/RadioCalibrationConfig.h
@@ -44,7 +44,7 @@ class RadioCalibrationConfig : public AP2ConfigWidget
 {
     Q_OBJECT
 
-    static const int RC_CHANNEL_PWM_MIN = 895; // Spektrum DX6i reports 898 on ch7 even though its 6 channels
+    static const int RC_CHANNEL_PWM_MIN = 850; // Spektrum DX6i reports 898 on ch7 even though its 6 channels
     static const int RC_CHANNEL_PWM_MAX = 2100;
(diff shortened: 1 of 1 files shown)


billbonney:
Is it sensible to lower the values, or reconfigure the end point sin the radio? I lowered to 895 since the DX6i was putting out a value at that value that could not be changed in the radio AFAIU. I'd like to keep it where it is?

lucasdemarchi:
@billbonney what's really the check for? AFAICS

In this RC I can't reconfigure it

billbonney:
Its the check that PWM signals are within the legal ranges before calibrating. What's the log say when you try to calibrate?

lucasdemarchi:
I don't have the log handy now, I can check it later.  But I have a dialog box with an error message saying the RC is not turned on.

billbonney:
That message might need to be update to be more informative

billbonney:
I've added the fix as is for now with commit 54df6c6. Though I'd still like to understand why you cannot change the limits on your 6XS

lucasdemarchi:
> I've added the fix as is for now with commit 54df6c6. Though I'd still like to understand why you cannot change the limits on your 6XS

Sorry for the delay in replying.  I looked through all the menus on the RC and haven't find a way to change min and max values. Last Thursday I found a menu in which apparently I can edit the values. I still need to test it.

However I still think we should not bother being strict in the values if there isn't a reason to.

PR #764, opened 2015-07-10
Sensor Setup Range Finder: Add BBB-PRU support

mirkix:
Add BBB-PRU support, according to commit in APM master
https://github.com/diydrones/ardupilot/pull/2519


billbonney:
Added with e193313 Thx :+1:

PR #765, opened 2015-07-14
Graph View: correct FMT header in plain-text export

peterbarker:


billbonney:
cherry-picked with 0ffc4b5 Thx :+1:

Issue #747, opened 2015-06-20
3DR Radio FW Update failing on linux


billbonney:
Flashing 3DR radio on linux fails
```
Flashing with firmware
./sik_uploader/firmware/1.9/radio~hm_trp.ihx
python: can't open file './sik_uploader/sik_uploader.py': [Errno 2] No such file or directory

Complete 2
```

martinjuju:
seems to work if the package is build from source. However, if install from .deb then the "sik_uploader" is not installed

PR #766, opened 2015-07-15
fixed radio uploader problem in Linux build

martinjuju:
this should fix the problem raised in issue #747 
add sik_uploader to the installation in Linux build.
apmplanner is looking for sik_uploader in usr/bin but the files were not installed previously.
This should fixed the problem by installing sik_uploader in /usr/bin


billbonney:
cherry-picked with 02c8069 :+1: Thx

PR #770, opened 2015-07-21
Camera mount RC channel selectors extended to 16 channels

froller:
Won't reset camera mount RC channels any more if they were set to 9 and above.

### DIFF
--- a/libs/opmapcontrol/src/core/providerstrings.cpp
+++ b/libs/opmapcontrol/src/core/providerstrings.cpp
@@ -42,10 +42,10 @@ namespace core {
 //        VersionGoogleLabels = "h@132";
 //        VersionGoogleTerrain = "t@125,r@132";
         // Google version strings
-        VersionGoogleMap = "m@301";
-        VersionGoogleSatellite = "171";
-        VersionGoogleLabels = "h@301";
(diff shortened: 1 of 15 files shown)


billbonney:
Hello to avoid the merge always of
`git pull --ff-only upstream master` and you will not get a merge.
```
git fetch upstream master
git rebase upstream/master
``` 
would just reapply your changes onto of master.

I cherry-pick changes onto master in anycase, so as long as it's not too far out of date, it good for me

Hope that helps :)

froller:
Thanks. After long years using SVN and P4 it's hard to understand Git dao. ;)
Some terms confuse me.

billbonney:
I'm ex-P4 and SVN, with some terrible ones in-between (Synergy aka Continuous, god help me) Git is really the easiest by far (except probably on a level with Mercurial)

I think the strangest concept in git and it's most power is the rebase, ie. the concept of rewinding the commits you have made to specific commit and then reapplying them in order to another branch (or same branch fast forwarded). This concept of rewind and fast-forward so all commits look like they are at the head revision. Makes for review simplicity.

http://stackoverflow.com/questions/457927/git-workflow-and-rebase-vs-merge-questions

We use the rebase model on this project (except for complicated merges)

Anyway, thanks for the changes.

PR #782, opened 2015-08-03
Added QDataStream missing include

Salamandar:
It does not compile without it.
The include may replace <QDataStream>, but it compiles+works with both.

### DIFF
--- a/src/ui/AP2DataPlotThread.cc
+++ b/src/ui/AP2DataPlotThread.cc
@@ -39,6 +39,7 @@ This file is part of the APM_PLANNER project
 #include <QSqlError>
 #include <QByteArray>
 #include <QDataStream>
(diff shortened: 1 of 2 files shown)


JoonasMelin:
This was also required for me to compile this on Arch.
Getting errors like: <code> apm_planner-2.0.16/src/ui/AP2DataPlotThread.cc:175:56: error: variable ‘QDataStream packetstream’ has initializer but incomplete type
                         QDataStream packetstream(packet);</code>
otherwise

midraed:
I also couldn't compile without changing those two files. 
Using Arch updated, wit Qt 5.5

billbonney:
closed with e65fbf7e1a3693d8bc63c893fae9450c9a713321

PR #783, opened 2015-08-04
Native Windows Builds 

dcarpy:
This should fix all issues when compiling Windows natively--at least it did for me.  :-)


billbonney:
What windows version is this for, and is it tested 64Bit and 32Bit?

dcarpy:
I think we should consider a x64 version.  Actually, only support a x64 version (if it builds)?

dcarpy:
Well, looking at it a little more closely it is going to be a lot more work than I thought to make it 64 bit.  New mingw and libs...maybe we branch and go from there if this is something you want me to explore?

dcarpy:
Scratch that Bill.  Thinking about it more, I forgot that ARM architecture like on the Pi is 32-bit.  Also, AP2 runs quick already and I really wonder how much gain we would get...

billbonney:
Don't forget we run 64bit for the OSX build, so the core supports it, the issue is updating the build env for windows.

dcarpy:
Hmm...good point.  Alright, should I write an enhancement bug and assign to myself?  Windows native, Windows cross-compile and Linux right?

dcarpy:
I got this bad boy to compile (native) 64-bit Windows.  It was a lot easier than I thought. I added MinGW-w64 and it compiled fine (debug and release).  Runs great--seemed faster but probably my imagination.  :-)

billbonney:
1a8fbdf would be a merge of upstream/master into your origin/master I would recommend that you keep your PRs in separate branch, since this makes the PR back to upstream/master easier to merge. 

In anywise, does these changes build for you with MSVC2012?

dcarpy:
Good advice on the branch deal--I'm more of a SVN kinda guy.  :-)

I did test these changes with MSVC2012:
Don't worry about dropping the force statement--must be a MSVC2010 thing.
Also, you fixed the CameraView already, so no need there either.
You completed the hard coded path to Git also...so really the SDL changes are needed.
Thanks Bill.

dcarpy:
BTW, fixing the conflict deal was more about me getting ready to check in the nsis scripts...but I'll do like you suggested and branch my repo instead.  Should I create a bug or enhancement request?

billbonney:
@dcarpy can you advise me if the PR should now be closed? Thx

dcarpy:
Yes please, you can close.

Issue #551, opened 2014-12-16
Add support for downloading behind a proxy server


billbonney:
Currently we do not support downloading when behind a proxy server. Need to add this

billbonney:
Change 71456bc435ede4c6cf804cd89c0017be98406ebd is partial use of Network Proxies. Still needs to be verified against use in FW donwload and other uses of QNetwork

PR #785, opened 2015-08-05
UI Framework: Enable use of Network Proxies

mdhorn:
Fixes Issue #551 Add support for downloading behind a proxy server

Enable the use of QNetworkProxyFactory to automatically
detect and use Network Proxies. This allows the downloading
of map data when behind a firewall.

Functionality can be disabled by setting environment variable
  IGNORE_PROXY

### DIFF
--- a/libs/opmapcontrol/src/core/opmaps.cpp
+++ b/libs/opmapcontrol/src/core/opmaps.cpp
@@ -118,7 +118,6 @@ namespace core {
                 connect(&network, SIGNAL(finished(QNetworkReply*)),
                         &q, SLOT(quit()));
                 connect(&tT, SIGNAL(timeout()), &q, SLOT(quit()));
-                network.setProxy(Proxy);
 #ifdef DEBUG_GMAPS
                 qDebug()<<"Try Tile from the Internet";
 #endif //DEBUG_GMAPS
--- a/libs/opmapcontrol/src/core/urlfactory.cpp
+++ b/libs/opmapcontrol/src/core/urlfactory.cpp
@@ -38,8 +38,6 @@ namespace core {
         /// timeout for map connections
         /// </summary>
 
-        Proxy.setType(QNetworkProxy::NoProxy);
-
         /// <summary>
         /// Gets or sets the value of the User-agent HTTP header.
         /// </summary>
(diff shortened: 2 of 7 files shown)


mdhorn:
Tested this change in an environment with a firewall and proxies configured as well as without a firewall and no proxies.

billbonney:
This only resolves proxies for Maps. AFAIU it doesn't fix the issue for Firmware Update, Software Update, TCP connections over a Proxy. (I'm sure there are some more things)

Maybe it would be an idea to break the issue into a numer of issues. Also it would be nice to see a config option in the APM Planner 2.0 Config view to manually set the proxy

Which OS was this changed tested against?

mdhorn:
Correct, this was specifically fixing the proxy usage with the Map download data.  However, this will enable proxies for any code using the QNetwork functions.

It was only tested under Linux (Ubuntu 14.04.2 LTS), but as it relies upon [systemProxyForQuery](http://doc.qt.io/qt-4.8/qnetworkproxyfactory.html#systemProxyForQuery) it should find the correct proxy settings for any Qt supported OS.

A configuration setting option was not created since it relies upon the OS system proxy settings, but I'm open to having a configuration that ignores the system settings if desired.

mdhorn:
Looking through the code it appears that both Firmware Update and Software Update are using QNetwork; they should follow the network proxy setting as well. 
Unfortunately I don't have hardware to allow me to test these cases.

I'm definitely open to making any changes to this patch set that are needed and testing what configuration I have available.

mdhorn:
Perhaps instead of enabled by default with an environment variable to disable, switching to using a configuration option to enable "Use OS Network Proxies" which is disabled by default?

billbonney:
That sounds like a good idea.

mdhorn:
Pushed a new version of the patch which replaced the environment variable with a configuration option.
Verified under ubuntu Linux; with and without firewall. Verified that changing the configuration option dynamically enables/disables the use of Network Proxies (for QNetwork code).

billbonney:
Thanks. I'll try and pull that tomorrow. Trying to fix a UDP issue introduced to OS X on 5.4.2. *sigh*

mdhorn:
I've made one more small change to ensure that the configuration box is unchecked if the proxy enablement fails.
Let me know if you have questions, concerns, or other desired changes to the patch.

billbonney:
rebased onto master with 71456bc435ede4c6cf804cd89c0017be98406ebd Thx :+1:

PR #810, opened 2015-09-29
UI Framework: Add missing QDataStream include for Qt5.0.0

gpaes:
This include is missing when compiling with Qt5.5.0.
I also tested successfully this patch with Qt5.4.1.


billbonney:
Thanks, this change is already in the qt55-wip branch which we are uisng to track Qt5.5 :+1:

PR #811, opened 2015-10-01
Fix race

tcanabrava:
I got into a recursion inside of MainWindow::instance() because the second call inside ApmFirmwareConfig::hideEvent was being called before the first call to MainWindow::instance() was finished. This happened because of a call to StackedWidget->setCurrentWidget hided the ApmFirmwareConfig, triggering the faulty call.


billbonney:
Thanks for the fix. That said its probably better solution to fix the port scanning to happen when the window is shown and remove the time start from the constructor entirely. What do you think?

tcanabrava:
Easy to do. will send a new patch in a few moments.

tcanabrava:
https://github.com/diydrones/apm_planner/pull/814

PR #819, opened 2015-10-05
Dont use QObjects in different threads (fix 802)

tcanabrava:
This code is experimental and by no means should be merged to
master: it still has bugs ( mainly, it broke the connect / disconnect
buttons ).

The code added in 20f7b8f to workaround a Qt bug also created a new
issue: it constantly changed the thread that an object was created
without paying attention about the signal / slot connection nor the
thread ownership.

This code moves all object creation from within the UDPLink object
inside the UDPLink thread, and the communication for the thread
contol is done via boolean flags (_restart, _reconnect, _finish, etc
) that the thread mainloop will certainly hit and do what is requested.

Tested and there is no more crash because of broken UDP content,
unfortunatelly it also broke the interface buttons for connection /
disconnection, but that's easily fixable.

Signed-off-by: Tomaz Canabrava <tomaz.canabrava@intel.com>

### DIFF
--- a/src/comm/UDPLink.cc
+++ b/src/comm/UDPLink.cc
@@ -37,6 +37,7 @@ This file is part of the QGROUNDCONTROL project
 #include <QTimer>
 #include <QList>
 #include <QMutexLocker>
+#include <QtNetwork/qhostaddress.h>
 #include <iostream>
 #include <QHostInfo>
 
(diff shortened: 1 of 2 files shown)


billbonney:
The changes in #819 works on OSX as I have just tested it
it’s using QMAKE_MAC_SDK = macosx10.11 (ie Xcode 7.0.1) toolchain

tcanabrava:
closed because I'm opening a new one that fixes things without breaking the connections.

PR #820, opened 2015-10-07
Dont use QObjects in different threads (fix 802)

tcanabrava:
The code added in 20f7b8f to workaround a Qt bug also created a new
issue: it constantly changed the thread that an object was created
without paying attention about the signal / slot connection nor the
thread ownership.

This code moves all object creation from within the UDPLink object
inside the UDPLink thread, and the communication for the thread
contol is done via boolean flags (_restart, _reconnect, _finish, etc
) that the thread mainloop will certainly hit and do what is requested.

Since a single approach is usually better than two different approaches
because it's easier to spot a bug and fix it, I removed the
BROKEN_SIGNAL define that was used to create two very different
approaches of the code and used only the code that should work for the
BROKEN_SIGNAL case, since it also should work for the non-broken signal
ones.

Tested and there is no more crash because of broken UDP content,
connection and disconnection of the UDP link works, no more random
crashes on linux because of threads and UDP connection.

Signed-off-by: Tomaz Canabrava <tomaz.canabrava@intel.com>

### DIFF
--- a/src/comm/UDPLink.cc
+++ b/src/comm/UDPLink.cc
@@ -28,15 +28,11 @@ This file is part of the QGROUNDCONTROL project
  *
  */
 #include <QtGlobal>
-#if QT_VERSION > 0x050401
-#define UDP_BROKEN_SIGNAL 1
-#else
-#define UDP_BROKEN_SIGNAL 0
-#endif
 
 #include <QTimer>
 #include <QList>
 #include <QMutexLocker>
+#include <QtNetwork/qhostaddress.h>
 #include <iostream>
 #include <QHostInfo>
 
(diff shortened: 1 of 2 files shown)


billbonney:
Thanks, this should fix https://github.com/diydrones/apm_planner/issues/802

When naming commits use the naming convention as specified in https://github.com/diydrones/apm_planner/blob/master/CONTRIBUTING.md

I'll rename this one on merge to master, but for future it would be something like

' UDP Link: Fixes#802 Dont use QObjects in different threads '

Thanks again for spending the time to fix this for the linux users on AP2 :+1: 

(It should be merged today, I am just testing with Qt5.5.0 as well)

billbonney:
Merged... Sorry forgot to close this.

PR #833, opened 2015-11-04
Fixes 808 - Segfault when loading mission twice

mrpilot2:


billbonney:
Cherry picked with 6ca8ce6023202e2ca718bd1086438458b511142d thx

PR #837, opened 2015-11-23
Graph View: Make errors show up like the mode does

Arne-W:
Errors will now show up in Graph View when enabled by ERR button below the graph.

I implemented this like the mode functionality and introduced a helper class providing a toString method for the Error codes. Perhaps this class should not be inside the AP2DataPlot2DModel file - but for the moment I found it acceptable.


src/ui/AP2DataPlot2DModel.cc
```diff
@@ -834,3 +881,198 @@ quint64 AP2DataPlot2DModel::getFirstIndex()
...
+{
+    return ((this->Timeus != lhs.Timeus) || (this->SubSys != lhs.SubSys) || (this->ErrorCode != lhs.ErrorCode));
+}
+
+QString ErrorType::toString()
+{
+    // SubSys ans ErrorCode interpretation was taken from
+    // Ardupilot/ArduCopter/defines.h
+
+    std::stringstream ss;
```
billbonney: Why use std:string here? QString support << 
see http://doc.qt.io/qt-5/qstring.html#operator-lt-lt

mrpilot2:
Converting these error codes to a string was already done here 5cd5deaf04e0e8d6dcde6498d6a5621bb9ede3d8 (although not used)

I would suggest to replace the old function with the new one, as it is related to the MAV type

src/ui/AP2DataPlot2DModel.cc
Arne-W: Ahh - okay I'm not very familiar with Qt but getting better :-). I'll make a fix for that.

Arne-W:
Do you mean replace the "content" of that function or remove that function an replace it with the new Error class?? I would prefer making an own class - perhaps with extended functionality if needed. It could be placed in a / the MAV namespace?!

mrpilot2:
Sorry for my late response

I agree with you - moving this code to it's own class would be the most desirable solution. Then the old code in ArduPilotMegaMAV can be deleted.

Arne-W:
No problem :-) 

Very good - I will change it in next few days. I think I will extend the class a little bit to reduce the pathological coupling in the AP2DataPlot2DModel::getErrorValues() method which knows about the internal representation of the ErrorType class.

billbonney:
The error code is related to APM only, and not other MAVs so it is associated with that vehicle type. It's the challenge with any higher order interpretations in the log files. Not sure what the solution is but being associated with the ArduPilotMegaMav at least indicates it's autopilot specific

Arne-W:
Okay I think I do understand. Perhaps it would be a solution to create an basic errortype with a generic toString method printing only the codes itself. This type would always be used for the generic handling of the errorcodes.
For each APM-Type (Rover, Copter, Plane....) we create a specialized class with an interpreting toString method, and a copy-constructor for creating the specialized class from the basic one.
This way you can take the basic error type cast it to the desired type an use the toString method.
I will change the implementation this way so you can have a look.

Arne-W:
Its done.
Take a look. I introduced a type hierarchy for the error type.

Arne-W:
After reviewing the code I am sure inheritance is an overkill at this place. So I removed the Inheritance an created a simpler formatter. less code - less complexity.
I think it is a better solution now.

billbonney:
Cherrypicked with 
* 8a29a0d Graph View: Removed errortype inheritance.
* 3c90c75 Graph View:Introduced Type herachy for ErrorTypes
* 3a264f1 Graph View: Moved ErrorType class to mav section.
* 31f6ed0 Graph View: Changed std::stringstream to QTextstream.
* c5a5689 Graph View: Make errors show up like the mode does

PR #841, opened 2015-11-26
Build: Qt 5.5.1 and Native Windows Builds

sutherlandm:
After some slight changes, mostly commenting out, things are running with Qt 5.5.1 (and Qt 5.4.2) and MSVC2013 on Windows, both 32 and 64 bit compiling. Unfortunately she didn't like CameraView, Joystick, and Magellan 3D mouse includes (among a few other small things).

I tested with SITL on both Windows 7 and a virtual Ubuntu 14.04 and didn't notice anything strange (other than no camera or joystick stuff). I do not currently have access to OSX. I also don't have physical hardware right now to test uploading/calibrating/etc. Most of the changes came from what QGC is doing as they use Qt 5.5.1 and MSVC2013.

I don't want to step on any toes though as it seems @dcarpy is already doing some work with native windows builds and MSVC2012. If we want to go this direction I will then see about fixing the CameraView and Joystick includes.

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -459,8 +459,8 @@ WindowsCrossBuild {
         $$BASEDIR/libs/lib/sdl/include \
 
         LIBS += \
-        -Llibs/lib/sdl/win32 \
-        -lSDL2.dll
+        -L$$BASEDIR/libs/lib/sdl/win32 \
+        -lSDL2
 }
 
 #
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -367,22 +367,25 @@ LinuxBuild : exists(/usr/local/lib/libxdrvlib.so) {
 }
 
 WindowsBuild {
-    message("Including support for Magellan 3DxWare")
+#    if defined(MOUSE_ENABLED_WIN)
+#    message("Including support for Magellan 3DxWare")
 
-    DEFINES += MOUSE_ENABLED_WIN
+#    DEFINES += MOUSE_ENABLED_WIN
 
-    INCLUDEPATH += libs/thirdParty/3DMouse/win
+#    INCLUDEPATH += libs/thirdParty/3DMouse/win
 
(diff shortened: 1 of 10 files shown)


billbonney:
Thx for PR. Here are some review comments.

Use defines to  remove sections of the build that are optional, since features build on other platforms ok. I'm thinking CAMERA_VIEW. 3D Mouse/Joystick also build when you have the correct dependencies included.

The update to the readme should not remove other options to build for windows. For example instructions to build using mingw should not be removed. Your changes enforce MSVC as the build environment, we need to keep mingw compatibility. We currently support cross compilation using mingw on linux on one of our build servers (though its broken at the moment). This PR would break that.

At the moment this PR cannot be merged, if you are interested in making it work across the platform and dependencies, I happy to collaborate and make it work

sutherlandm:
Oops, sorry was going to talk to you first on how to properly do this, still new to this forking/pull requesting business. 

Indeed I'll add the defines where needed, was more of a quick (and dirty) way of seeing if things would even work first. The issue I have is MSVC throws *LNK2019 unresolved external symbol* error when say the joystick widget is called. I guess this is something for Gitter or raise an issue though. 

I didn't realize the cross-compile was that important and I figured it would have taken me longer then it did to get it working. My bad, hopefully this is more acceptable.

billbonney:
Thanks, that would be awesome :+1:  

I'm not sure if newer version of MSVC are free, hence I'd line to keep mingw as option, and it allows cross-compilation on linux. It would also be nice to support a 64bit build for windows. but we still need 32bit to work for RPi and other 32Bit platforms. 

I generally test 32Bit on an old linux laptop, but main development for me is OSX. I do need to get an automated windows/linux build working again, so any work you do to that end (i.e. MSVC or mingw windows build) will be really helpful. I'm have MSVC2012 installed on a WIN7 laptop, but if I can upgrade for free I can do that.

sutherlandm:
Working on a better implementation, closing for now.

Issue #40, opened 2013-08-16
Fix compiler warnings


malcom2073:
Project should be changed to -Wall -Werror, and all compile issues fixed.

Arne-W:
Closing this as the only warnings left are not in APM planner code.

PR #842, opened 2015-11-28
Fix compiler warnings

mrpilot2:
Fixed most of the compiler warnings occurred on my Ubuntu 14.04 using Qt 5.3.0

Fixes issue #40

### DIFF
--- a/QsLog/QsLogDestFile.cpp
+++ b/QsLog/QsLogDestFile.cpp
@@ -122,10 +122,10 @@ void QsLogging::SizeRotationStrategy::setBackupCount(int backups)
 
 
 QsLogging::LaunchRotationStrategy::LaunchRotationStrategy()
-    : mCurrentSizeInBytes(0)
+    : mFirstFile(true)
+    , mCurrentSizeInBytes(0)
     , mMaxSizeInBytes(0)
     , mBackupsCount(0)
-    , mFirstFile(true)
 {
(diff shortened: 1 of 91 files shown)


billbonney:
I'ver cherry-picked all these as follows
* e673c46 Fix compiler warnings: Qt Ui Invalid z-order assignment (issue #40)
* 7919d78 Fix compiler warnings: removed set but unused variables (issue #40)
* 35a4097 Fix compiler warning: unused parameter (issue #40)
* 0795d16 Fix compiler warning: unused parameter (issue #40)
* 7000f6e Fix compiler warning: unused parameter in libs (issue #40)
* 6c9f161 Fix compiler warning: enumeration value not handled in switch (issue #40)
* 84ac262 Fix compiler warning: deleting object of polymorphic class type which has non-virtual destructor (issue #40)
* 28fab89 Fix compiler warnings: case value 38 not in enumerated type QVariant::Type and comparison between enum QVariant::Type and enum QMetaType::Type
* 71c7fc4 Fix compiler warning: ignored qualifiers (issue #40)
* 20d41e1 Fix compiler warning: suggest explicit braces to avoid ambiguous else (issue #40)
* db25ab1 Fix compiler warning unused but set variable and unused variable (issue #40)
* 46f0d04 Fix compiler warning initalization sequence (issue #40)

PR #846, opened 2015-12-01
Build: Windows MSVC2013 and MinGW work

sutherlandm:
Small changes to compile with MSVC2013 (32 and 64 bit) and MinGW (32 bit). Included with 32 and 64 bit SDL libs, it would cause a link error with 64-bit MSVC2013. Also updated the Window README section to outline the install, setup, and compile.


billbonney:
This PR was cherry-picked into msvc-wip (we should work off that branch until we get the build working) So raise further PRs to that.

I also 'fixed up' the 'add back MSVC2012 support' since it wouldn't compile on OSX

sutherlandm:
The 'add back MSVC2012 support' was from the old branch qt551-winmsvc where as I redid things on qt551-wip. Regardless I'll work off the one above.

billbonney:
This has been merged to master with 69161a75d18cbe0b16518bc14adad22b49c21a62

billbonney:
PS: still add further changes to mdvc-wip for building the installer etc... I'll start testing and integrating that into my TeamCity build server now.

PR #848, opened 2015-12-07
Build: Windows compile work

sutherlandm:
Change check from Q_OS_Win to looking if _MSC_VER is defined as the former causes build failing with MinGW. Also remove MSVC2010 from supported Windows compilers.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -49,7 +49,7 @@ linux-g++-64 {
          DEFINES += Q_UBUNTU
     }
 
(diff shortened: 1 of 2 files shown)


sutherlandm:
Actually @billbonney, if you just want to change (Q_OS_WIN) to (_MSC_VER) and commit, I think it would get the same result with nicer looking code.

billbonney:
cp with f23fc665705d7d1c92723a15ae772f6000a763b0 & a662992ad1edde6175bcb6c0982f4572d403cf6f

PR #849, opened 2015-12-07
Fix 834

mrpilot2:
I've created a unit test that reproduces the segfault, no matter if filte1.4, or flite2.0 is installed on my Ubuntu 14.04

The unit test works fine for flite1.4 and flite2.0, if commit  5f0890a72bf227fc010c68517fe91b44efa850e1 is reverted


billbonney:
Thanks for finding a solution. I can't merge this commit until the fix is against current master. I know I need to pull in the other two PRs you have, but this PR is essentially everything is an 'omnibus' PR.

I can cherry-pick dc60229, but it much easier to merger PRs against the current master without the 'noise' of other PRs

mrpilot2:
Sorry for that, but I wanted to cover my changes with a unit test, that's why I used my master branch.

billbonney:
No worries. I'll look at the unit test you added again, the problem was I had some OSX build issue, and I hadn't had time to resolve (I'll try and find some time to fix that as it's a good PR)

billbonney:
cherry-picked fix with 6f4ec1fbd97de2527fd6fb40afc46d62304e550e

That still leaves the unit test case pr and then the two test cases

PR #850, opened 2015-12-09
Main Window: Consistent style of main toolbar

sutherlandm:
The main window toolbar would follow the style of host OS (bright blue with Windows 7, or two tone gray on Ubuntu for example). Will now always have black background (to match the APM toolbar) and white text with either indoor or outdoor style/color. 'Inner' menus still follow the dark/light style.

### DIFF
--- a/files/styles/style-indoor.css
+++ b/files/styles/style-indoor.css
@@ -345,6 +345,14 @@ QMenu::separator {
      margin-right: 5px;
  }
 
+QMenuBar {
+    background: black;
+}
+
+QMenuBar::item {
+    background: black;
+}
+
 QMenuBar::item:disabled {
     border: none;
     background: none;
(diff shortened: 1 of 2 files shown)


billbonney:
Looks a bit odd with the menu drop down not being in sync with the color 
This change has no impact on OSX
menucolor

sutherlandm:
I was on the fence about how it looked, changed it so it will always look like below. Not sure why OSX would not work, I'll look into it.  
colour_fix

sutherlandm:
Ah right. Ok well it is up to you if you want to change it or close this to just leave things as is for now, my feelings won't be hurt. Would it be worth exploring the option of no menu bar like Mission Planner?

billbonney:
Closed: For now we are keeping it as is. Thanks for the proposal though 👍

PR #853, opened 2015-12-14
Graph: tooltip will show coordinates, milliseconds better

AndKe:
Lat,Lon was displayed with only two decimals.
ms was displayed with exponent (giving no clue about seconds)


billbonney:
cherry-picked with a59c555

PR #855, opened 2015-12-20
Graph View: Bugfix: Log Export not complete. Fixes #796

### LINKED ISSUE #796: Log Export not complete
**keohr** see http://ardupilot.com/forum/viewtopic.php?f=82&t=13600
With recent compiled APM Planner 2.0 (2015-08-27) I get some inconsistent / missing log lines when exporting from bin -> log.

1. 
FMT, 133, 35, RCIN, Ihhhhhhhhhhhhhh, TimeMS,C1,C2,C3,C4,C5,C6,C7,C8,C9,C10,C11,C12,C13,C14
Defined are C1 to C14 but only C1 to C12 are exported:
RCIN, 75375, 1490, 1526, 982, 1520, 1166, 1494, 982, 982, 874, 874, 874, 874
(in this case LogAnalyzer reports BAD LINE:)

2.
The MSG line(s) is -not always- exported:
So sometimes these lines are missing in the export:
FMT, 132, 67, MSG, Z, Message
and for example:
MSG, ArduCopter V3.2.1 (36b405fb)
MSG, PX4: ce602658 NuttX: 475b8c15
MSG, PX4v2 00320030 3034510A 38353334
MSG, Frame: QUAD

3.
The Parameters 
PARM, THR_MIN, 130
PARM, THR_MAX, 1000
and perhaps more are not exported
(LogAnalyzer exits with error)

Hope this will help to improve APM Planner.

Arne-W:
Valuepairs were only stored if size was bigger than 1 but all non empty pairs need to be stored.
This fixes point 2 - The MSG line(s) is -not always- exported - mentioned in #796 

I was not able to reproduce the two other points 
* 1. - Defined are C1 to C14 but only C1 to C12 are exported
* 3. - Missing Parameters

Tested with ArduCopter V3.2.1

### DIFF
--- a/src/ui/AP2DataPlotThread.cc
+++ b/src/ui/AP2DataPlotThread.cc
@@ -342,7 +342,7 @@ void AP2DataPlotThread::loadBinaryLog(QFile &logfile)
                                     QLOG_DEBUG() << "AP2DataPlotThread::run(): ERROR UNKNOWN DATA TYPE" << typeCode;
                                 }
                             }
-                            if (valuepairlist.size() > 1)
+                            if (valuepairlist.size() >= 1)
                             {
                                 if (!m_dataModel->addRow(name,valuepairlist,index))
                                 {


Arne-W:
Today I did some testing with ArduCopter V3.3 and it seems that this issue is only related to ArduCopter V3.2.1 logs because since V3.3 all messages have a timestamp which extends their length by one.

billbonney:
cherry-picked with e13097b

Issue #708, opened 2015-05-14
"export log" is extremely slow.


AndKe:
a 25MB .bin log is opened within ~15 seconds.  on opening, it's basically converting to near .log format.
exporting the very same log to a .log , takes ~20 minutes
(the task takes 2% CPU , the drive is a fast SSD)

PR #856, opened 2015-12-20
Graph View: Speed up binary log parsing and log exporting.

Arne-W:
While making tests for the last bugfix I recognized that the log export in Graph View was extremely slow. The root cause was excessive database access while exporting the logs.
I reduced the DB access by caching some data.
Furthermore I commented the log line formatting while parsing binary logs (gave a little speedup too).
Vehicle type detection is now done comparing values of the valuepairlist directly.


Arne-W:
Today I found that there is already an issue for this #708

billbonney:
cp with a0578ba & ad4fc58

PR #863, opened 2015-12-31
Graph View: Double the speed of binary log parsing.

Arne-W:
* Doubled the number of inserts per sec. from 20000 to 50000 by using prepared insert queries.
* Changed all query pointers to smart pointers
* cleanup CTOR and removed dead code

### DIFF
--- a/QGCSetup.pri
+++ b/QGCSetup.pri
@@ -135,6 +135,27 @@ WindowsBuild {
     COPY_FILE_LIST = \
         $$BASEDIR_WIN\\libs\\thirdParty\\libxbee\\lib\\libxbee.dll \
 
+        # [TODO] Do we need to remove the 3 line below for windows?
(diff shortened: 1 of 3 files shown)


billbonney:
I merging this in as in, but the comments can be applied with a new PR as it's only a code review issue.

Arne-W:
OK - Done. Will you pick it from this PR or shall I create a new one??
By the way - I am slightly confused where the commit from dcarpy comes from. Do you have any idea - I did not pick it by intention!?

billbonney:
I can cherry-pick the update, so no worries. 

The commit from dcarpy come in since i pushed it to master. you based your changes on that commit, then I removed it from master and forced pushed an update to master. the change is still in your branch, but not master, so now it shows.
<TLDR> Don't worry about it ;)

Arne-W:
Ahh okay I understand. Thanks for the explanation.

billbonney:
cherry-picked with 36cd961thx :+1:

PR #865, opened 2016-01-02
MSVC 2015 Runtimes and NSIS Update

dcarpy:
Added Microsoft Visual C++ 2015 Redistributable (x64) and Microsoft Visual C++ 2015 Redistributable (x86).  Also, made changes to the NSIS script for native Windows build.  This should help with Windows 10 and Windows 32-bit installs, however, I don not have either to test changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ user_config.pri
 *.opensdf
 apmplanner2.xcodeproj/
 GeneratedFiles/
+deploy/*.exe
 deploy/*.list
 *.exe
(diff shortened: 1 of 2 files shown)


billbonney:
OK, Sorry. I thought I had replied to the pull, until we test, I think we should just keep this PR here? Comments?

dcarpy:
I did test, just not with Windows 10 is what I should have said.

billbonney:
@dcarpy I guess we can close this? reopen if you disagree

PR #867, opened 2016-01-03
Graph View: Speedup export and sort Fixes #864

### LINKED ISSUE #864: Graph speedup.
This commit works great:
https://github.com/diydrones/apm_planner/commit/1092d9af4ed40a5f13da0ffdde388820d333b3a6

please speed up the sort feature (which still takes the same time as before patch.)
and .log export.

If a certain log loaded in 21s before, filtered using "show sort"  in 21s before  , it takes now 8s to load, but still 21s to filter.

Arne-W:
I migrated the new prefetch methods I introduced in a0578ba28b80cfafd021438cbbfb2f00b51aa911 into the old data method which will do the caching now. All functions using the data method will now benefit from the caching, not only the export feature.
Through using prepared select queries the log export got another boost. Exporting 36Mb logfile takes now round about 22 seconds.
Furthermore I tried to optimize the sorting feature. Due to a lack of knowledge about the internals of the QSortFilterProxyModel I did some erratic testing. The result can be seen below. I don't know if this is good QT style but on my Ubuntu 15.04 this solution gave the best performace on huge logs. My testing log was 15Mb binary one with round about 520000 loglines. All filter actions are now done in less than one second.
So test and have fun!!

### DIFF
--- a/src/ui/AP2DataPlot2D.cpp
+++ b/src/ui/AP2DataPlot2D.cpp
@@ -1466,6 +1466,8 @@ void AP2DataPlot2D::exportButtonClicked()
 }
 void AP2DataPlot2D::exportDialogAccepted()
 {
+    QElapsedTimer timer1;
+    timer1.start();
     QFileDialog *dialog = qobject_cast<QFileDialog*>(sender());
     if (!dialog)
     {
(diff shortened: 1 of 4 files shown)


AndKe:
opened 77MB .bin in 48seconds    , and any filter is applied within 2s.  
That's a great achievement ! - it's a pleasure to work with logs now.  - well done ! :)

opened a 2.4MB .bin, exported to .log , with previous build, 22sec.  now exported to .log in 12seconds
clearly improved :)
Thank you.

Arne-W:
Hi AndKe,
nice to hear that you are satisfied with my solution :-)
The only thing that makes me wonder are the times you measured. On my machine importing a 15 MB binary log takes about 11 sec and exporting to ASCII log takes 22 secs.
You say importing a 70Mb bin log takes 48 sec wich matches my expectation. But exporting a 2.4Mb bin takes 12 sec. This is too long I think cause in this time I can Export a 8Mb file.
I would expect a time below 5sec for your Export.
Gould you please verify your measuring. If it leads to the same results it would be nice if you could send me the 2.4Mb log for testing.
Thanks

AndKe:
I too noticed some discrepancies between the time we mentioned, but decided not to nag about it. I am glad you care about such details anyway.
The attached 2.4MB lo loads in about 1,5sec (stopwatch)  - and exports in 12.8s this time (stopwatch as always)
https://www.dropbox.com/s/xarjdb5q5mgnqts/HEXA%2Cnormal%20wind.BIN?dl=0

it should be noted that I am testing this on master from yesterday with PR 867 , in case you had some other export optimizations in another PR...

Arne-W:
Will check it tomorrow evening

Arne-W:
I did some testing this evening and my results are like I expected. From the log:

INFO 2016-01-06T22:02:47.694 Plot Log loading took 1.848 seconds - 2537922 of 2537922 bytes used
DEBUG 2016-01-06T22:03:51.124 Log export took  3426 ms

My system is a Kubuntu 15.10 and the machine is a Core I5-2520M 2,5 GHz with 4GB Ram.
So it seems that the time differences are not not related to your log file. Perhaps its hardware related but your loading time is nearly the same as mine. Why can you load that fast but not export - strange.

billbonney:
cherry-picked with 126f7fc & 72c65dd

AndKe:
@Arne-W  my tests are done on Ubuntu 15.10 , with "won't brag hardware"  - but I often use it for map stiching, so you can imagine.
Just tried the master with fixes that were pulled today, and I have the same results - right now  I guess it may be due to the fact I am using qmake 5.4.2   , qt5.5 is installed in /opt, but I forgot if it only needed to be in path or needed a variable as well.  :)
Anyway - the most important speedups work great.

Issue #868, opened 2016-01-05
Graph: IMU.Acc? data not plotted, visible in grid.


AndKe:
please see this post: http://ardupilot.com/forum/viewtopic.php?t=14691
The Acc data is not plotted on current master.
and for what it's worth: mavlogdump handles the logfile just fine.

Arne-W:
On Linux I can not reproduce this (Kubuntu 15.10)

mboland:
This seems intermittent, as I downloaded some logs and the IMU data did display properly.
Same machine, same copter, next day flights and the IMU data did not display again.
Mac OS X 10.11.2

Arne-W:
@mboland could you test if the problem persists with the actual master and if it is still there upload a log witch shows the problem? I could test again then.

AndKe:
@Arne-W   - the log is here: http://ardupilot.com/forum/download/file.php?id=4850&sid=6f6ba2b99101805859f769fd34d75567  - and it's still an issue in current master - I had no seeing those data, from that particular file plotted, ever - tried at least 6 times.

mboland:
Sorry for dumb question but where would I download Master from?

AndKe:
@mboland right here, git clone https://github.com/diydrones/apm_planner.git  
- in case you look for compiled binary (may I guess that's the question?) 
it's supposed to be here, but something apparently broke and the build server makes only mac versions ?

billbonney:
prebuilt here for OSX http://firmware.diydrones.com/Tools/APMPlanner/daily/2016-01-06/

Arne-W:
@mboland - Just tested your log and you are right - no IMU data. The main problem is that your log data is corrupted! While loading your log the APM-Planner-logging says:
DEBUG 2016-01-10T12:59:46.617 AP2DataPlotThread::run(): Non packet bytes found in log file 1606 bytes filtered out. This may be a corrupt log 
DEBUG 2016-01-10T12:59:46.617 AP2DataPlotThread::run(): Non packet bytes found in log file 3526 bytes filtered out. This may be a corrupt log 

Opening your log with an HEX-editor shows a corrupt end (only FF FF FF FF....) and the APM-Planner table view shows "nan" - Not A Number in the last table line.
![log_err](https://cloud.githubusercontent.com/assets/7621911/12221447/b4ea0968-b79b-11e5-95b7-c5acdace8491.png)
[...]

PR #877, opened 2016-01-16
Graph View: Create more detailed error message on log loading

Arne-W:
As discussed in #868 here is my solution for providing more details about errors raised during log parsing. Hope you like it. If you have any Ideas for better messages or anything else - just let me know.
By the way - I was not able to resize the message box setFixedSize() did not work as expected. It would be good to increase the width.
This is what it looks like now:
new_error_popup1
new_error_popup2
new_error_popup3

By the way during developing I foud out that most of my APM 3.2.1 binary logs have an error during format parsing.


Arne-W:
@billbonney I don't want to be impatient, but the Tlog parsing crashes on current master and I made a fix for that and some other issues on Tlog parsing :-)
But it needs this PR for the better error handling. It would be nice if you could have a look.

billbonney:
done :+1:

Arne-W:
Thanks :+1:

PR #878, opened 2016-01-20
Graph View: Crash fixing and restructuring of TLOG parsing

Arne-W:
While doing some test I noticed that Tlog parsing is very slow and fails sometimes which leads to a segfault due to null-pointer access. So I fixed this and restructured the tlog parsing. This is the result.

* Heavy restructuring of TLOG parsing:
  - removed several memleaks and switched to shared pointer
  - removed time handling cause using it as index lead to DB errors
  - improoved errorhandling uses m_plotState now
  - introduced filtering of "EMPTY" messages as they have no format info

* Moved metatype registration to AP2DataPlotThread class
* Fixed null pointer access in AP2DataPlot2D::threadError


billbonney:
Thx :+1: These changes are great.

Arne-W:
Nice to hear that you like it :-) - and please let me know if there are any code style or QT issues in my code as I am working with QT for the first time and still learning.

PR #879, opened 2016-01-22
Graph View: Support for plotting data vs. time if all data has microsec timestamp

Arne-W:
All data in Graph View is plotted against an artificial data index. This is a robust solution as not all logs support a valid time-line but it makes data matching against flight time impossible.

Since arducopter 3.3 all bin logs have a microsec time stamp on every value. So I thought it should be possible to plot the data vs. time and not only vs. index if the log has a valid time-line.
After a lot of work (you know the first feature prototype takes only 20% of the time the whole feature takes) I have made it. if there is a valid time-line you can switch between time or index on x axis, if not the check-box for switching is hidden.

I made 2 screenshots where you can see nearly the same data. First one shows data vs. time and the second shows data vs. index. The plot vs. time looks far better and more natural.
timeline
timeline2

Changes overview:
* Added checkbox to switch between data vs. time and data vs. index
* Added several helper functions to avoid code doubling
* Fixed crash on load log canceling
* Removed some mem leaks
* Extended AP2DataPlot2DModel to support time indexes
* Cleanup and commenting

### DIFF
--- a/src/ui/AP2DataPlot2D.cpp
+++ b/src/ui/AP2DataPlot2D.cpp
@@ -58,17 +58,16 @@ AP2DataPlot2D::AP2DataPlot2D(QWidget *parent,bool isIndependant) : QWidget(paren
     m_plot(NULL),
     m_wideAxisRect(NULL),
     m_logLoaderThread(NULL),
-    m_model(NULL),
     m_logLoaded(false),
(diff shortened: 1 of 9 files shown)


billbonney:
:+1: Awesome. I'll pull that in later today. That's a really needed feature.

PR #885, opened 2016-01-29
Graph View: Some small bugfixes

Arne-W:
Fixes:
* Table View does not start at 0 
* Table View sorting did not work correctly when used for the first time
* Graph selection will be saved now when switching x-Axis to timeline


billbonney:
Cherry-pick with b0d0fa0 bca0d75  c2d4e15

PR #886, opened 2016-02-03
Graph View: Added Event plotting and unified handling of messages Fixes #870

### LINKED ISSUE #870: Graph: not all Events/error are marked in the log
please see attached log.
only the first event/log are graphed in the log, but there are many more.

[10.BIN.zip](https://github.com/diydrones/apm_planner/files/78516/10.BIN.zip)

Arne-W:
I completed my work on ' show EV' check-box. It now has the expected functionality.

This is what I have  done:
* Added functionality to 'show EV' check-box.
* Unified handling of 'MODE,ERR,EV' messages by using simple inheritance
  * Added abstract base class for all messages
  * Implemented special types for 'MODE, ERR, EV' messages
* Moved old Mode enums into new namespace
* Colourized plotted arrows
* Verified all printed messages against apm code. MODE for all types (copter, plane, rover), ERR and EV for copter only.

So sorry I had to remove that "TableView does not start at 0" Bugfix because it destroys the relation between the table and the graph. The fixing of this issue seems to be too big as the starting at 0 feature is more 'cosmetic' style.
* Removed Bugfix for 'Graph View: Fixed TableView does not start at 0' -> does not work - destroys relation between graph and table.

Looks like this now:
ev-work

### DIFF
--- a/src/uas/ApmUiHelpers.cc
+++ b/src/uas/ApmUiHelpers.cc
@@ -5,12 +5,13 @@
 void ApmUiHelpers::addRoverModes(QComboBox* comboBox)
 {
     comboBox->clear();
-    for (int count = 0; count < ApmRover::modeCount; count++){
-        QString mode = ApmRover::stringForMode(count);
-        if (!(mode.contains("Undefined", Qt::CaseInsensitive)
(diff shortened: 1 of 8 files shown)


billbonney:
cherry-picked with 0b579a8 :+1:

Issue #209, opened 2014-02-08
tool tips only half visible


mboland:
The tool tips on the serial connection window are cut in half vertically so only the top half of the text is visible.
e232e624 indoor Theme
2/3rds of the text is visible in Outdoor Theme
Mac OS 10.9.1

![screen shot 2014-02-09 at 6 43 47 am](https://f.cloud.github.com/assets/3221133/2118378/8f461bbe-9102-11e3-8d61-6a9815ea7b37.png)

Also in Communications Console but only on the serial port drop down
![screen shot 2014-02-09 at 6 52 27 am](https://f.cloud.github.com/assets/3221133/2118384/6afeed02-9103-11e3-8473-4484ca1a8e0e.png)

The other items display correctly
![screen shot 2014-02-09 at 6 52 45 am](https://f.cloud.github.com/assets/3221133/2118386/7c62882e-9103-11e3-96e5-f4662508296e.png)

and if the cursor is moved to the serial port drop down quickly enough it also displays correctly.
But move it out and back in, and the text displays cut off again.
![screen shot 2014-02-09 at 6 52 53 am](https://f.cloud.github.com/assets/3221133/2118388/ab3d6998-9103-11e3-8161-d0219dbe1eca.png)

PR #888, opened 2016-02-04
QT Styles: QToolTip visibility

ustramooner:
In indoor-styles, some qtooltips are not fully visible
fixes bug #209
tested on linux and mac


billbonney:
cp with 1c516c6

PR #890, opened 2016-02-04
Quick Status View: status tool tips

ustramooner:
Some parameter names are cutt off - show tooltips


billbonney:
cp with 2311a12 :+1:

PR #891, opened 2016-02-04
Info View Raw Status: show table immediately

ustramooner:
table view takes 2 seconds to display which can be confusing


ustramooner:
forgot to mention that this is actually also solves a bug where on linux the status view doesn't even show up. but has the 'added' benefit of showing the table immediately instead of waiting for the timer

billbonney:
cherry-pikced  with 207edcc

PR #892, opened 2016-02-04
UAS Quick View: status tool tips

ustramooner:
tooltips for quick view


billbonney:
I've rejected this change as it doesn't make sense in the UI for tooltips here

PR #893, opened 2016-02-05
EKF Monitor, Vibration Monitor: fix for ubuntu 14.04

ustramooner:
QtQuick.Controls 1.1
depends on qtdeclarative5-controls-plugin

note:
* not sure how to deal with platforms that have QtQuick.Controls 1.2 ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Be sure to run apt-get update first
 sudo apt-get update
 sudo apt-get install qt5-qmake qt5-default \
   qtscript5-dev libqt5webkit5-dev libqt5serialport5-dev \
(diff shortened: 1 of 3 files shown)


billbonney:
What version of Qt is this change for?

ustramooner:
@billbonney 5.2 - which i thought would have had a higher version for QtQuick.Controls. sorry, i'm not really very familiar with Qt - but this helped me to get those widgets working

QMake version 3.0
Using Qt version 5.2.1 in /usr/lib/x86_64-linux-gnu

billbonney:
You need to install Qt5.4.2 (we don't support Qt5.2 anymore)

Issue #472, opened 2014-09-26
Camera Gimbal View: Fix comboBox not being set as disabled


billbonney:
When you reassign an output channel from e.g PAN to TILT the old channel assignment is disconnected. The issue is that the UI is not updating to reflect this new state

see discussion here https://github.com/diydrones/apm_planner/issues/440

billbonney:
I need to verify this change since it's quite the UI and conditions. Hopefully over the next couple of days.

PR #894, opened 2016-02-05
Camera Gimbal Config: update interface when disabling other outputs

ustramooner:
when an output is chosen which is already in use the interface needs to disable the old output
this PR fixes bug #472


billbonney:
Thanks, merged with f862eac 👍

PR #896, opened 2016-02-05
Logging/Graph usability

ustramooner:
Various improvements to the log page:

* ability to jump to the table location when clicking on the graph
* shows a red-line at the currently selected table row
* remembers previous settings
* adds an 'ok' button to the warning screen

### DIFF
--- a/src/ui/AP2DataPlot2D.cpp
+++ b/src/ui/AP2DataPlot2D.cpp
@@ -362,7 +362,7 @@ void AP2DataPlot2D::verticalScrollMoved(int value)
 }
 
 void AP2DataPlot2D::plotDoubleClick(QMouseEvent * evt){
-    if (!ui.jumpToLocationCheckBox->isChecked())
+    if (!ui.jumpToLocationCheckBox->isChecked() || m_useTimeOnX )
     {
         return;
     }
@@ -1818,6 +1818,9 @@ void AP2DataPlot2D::setupXAxisAndScroller()
 
 void AP2DataPlot2D::plotCurrentIndex(int index)
 {
+    if ( m_useTimeOnX )
+        return;
+
     QLOG_DEBUG() << index;
     m_timeLine->start->setCoords(index, 999999);
     m_timeLine->end->setCoords(index, -999999);
@@ -1825,6 +1828,8 @@ void AP2DataPlot2D::plotCurrentIndex(int index)
 }
 void AP2DataPlot2D::insertCurrentIndex()
 {
+    if ( m_useTimeOnX )
+        return;
     QCPAxis *xAxis = m_wideAxisRect->axis(QCPAxis::atBottom);
     QCPAxis *yAxis = m_wideAxisRect->axis(QCPAxis::atLeft);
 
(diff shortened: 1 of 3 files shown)


Arne-W:
Very cool feature :+1:  but your branch does not build. It looks like the header entries for `void plotCurrentIndex(int index);` and `void insertCurrentIndex();` are missing.

Funny thing with the "missing ok Button" - on my system it does not make any difference there is always an OK button. But I think its better to create one to be sure its really there!

billbonney:
Please can fix this against current master and I'll merge the PR

ustramooner:
i did a little refactor which i forgot to commit. history rewritten

billbonney:
Thanks :+1:  I would suggest using git rebase your changes on master (instead of merge from master) 

Also I would suggest git rebase -i HEAD~<num_of_commits_rebase> to cleanup the history.

I can fix it this time if needed

Arne-W:
Today I did some testing and found two issues. 
First one is easy to fix because it just deals with the Index offset between graph and table. I could be possible that this issue depends on my reverted bugfix for the table view index did not start at 0.

Changing these lines should fix it:
`QModelIndex sourceIndex = m_tableModel->index(position, 0);` 
to
 `QModelIndex sourceIndex = m_tableModel->index(position - min, 0)`
at line 394 and 403 in AP2DataPlot2D.cpp
and adding this line
`index += m_tableModel->getFirstIndex();`
to the `void AP2DataPlot2D::plotCurrentIndex(int index)` method also in AP2DataPlot2D.cpp.

The second issue is that this feature does not work when using time on X-axis and I think this is a little bit more complicated due to a missing relation between time and table-index in the data-model.
But I think it should be possible to create such a relation while inserting the data into the model.
However I think we should extend this feature to work with time X-axis too.

ustramooner:
Hi Arne,

What's the situation where this happens? I'm thinking wrap-around logs?
I would think the bug would then also exist for all situations
where plotCurrentIndex is called?

Second issue: what kind of data would you want to display time on the
X-axis?

Ben

Arne-W:
Ah - okay my fault - perhaps first I should describe what the first issue is :-)
The red "line cursor" does not exactly match the the position in graph when clicking into table and when double click into the graph it hits not exactly the table line. To test just zoom insanely deep into the graph and try to position the "red line cursor" exaxcly over one of the text arrows for Mode, Err or Ev. You will see that there is an offset.

The second issue can only be verified with a binary log (flash log) from a arducopter 3.3 Firmware. If you load such a log there will be a button to switch the x-axis from index view to time view (see #879). When switched to time axis the index you get on a double click is not the table index any more but the time stamp of a special row. So we have to create a function which makes it possible to convert a time stamp into a table index.

ustramooner:
@Arne-W i've fixed that - you're right it wasn't accurate when you zoomed in. 

i'm not sure how to deal with the second issue sorry - i've disabled the red line/etc functionality when in this mode, unless someone else can solve this?

Arne-W:
Works like a charm :+1:  
I have some Ideas about creating some methods in the datamodel which can be used to convert the Table-index into a time stamp and vice versa. So enabling this feature for time axis can be done in a second step.

billbonney:
Cherry-picked , thx :+1:

PR #902, opened 2016-02-19
Config: accelerometer calibration fix

AndKe:
*It did not work (at least for AC3.3.2)  because "place vehicle"
messages are MAV_SEVERITY_ERROR (3)  - not CRITICAL (2)
* hiding the useless "initializing APM" messages not to confuse user
* handling if the actual command 241 should fail

### DIFF
--- a/src/ui/configuration/AccelCalibrationConfig.cc
+++ b/src/ui/configuration/AccelCalibrationConfig.cc
@@ -174,10 +174,10 @@ void AccelCalibrationConfig::uasTextMessageReceived(int uasid, int componentid, 
     Q_UNUSED(uasid);
     Q_UNUSED(componentid);
 
-    if (severity <= MAV_SEVERITY_CRITICAL)
+    if (severity <= MAV_SEVERITY_ERROR)
     {
         //This is a calibration instruction
-        if (!m_isCalibrating || text.startsWith("PreArm:") || text.startsWith("EKF") || text.startsWith("Arm"))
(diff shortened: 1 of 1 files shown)


billbonney:
The problem is a part fix slipped into the code 4a4bda6544a52de5f4295ade1b9f603fae8e658a

The issue is that the SEVERITY has been normalized to be MAVLink standard on the autopilot. AC3.4 AP3.4 > and this is different that AC3.3

I have  local branch that extends https://github.com/billbonney/apm_planner/commit/c223ab7b8c7b8eafb3f2f19967c3fa8dc422544f to allow a fix to be enabled/disabled as an option, it's just not pushed. I see if i can find time to do that this weekend

see https://github.com/billbonney/apm_planner/commit/c223ab7b8c7b8eafb3f2f19967c3fa8dc422544f for the old changes (not complete)

billbonney:
Scrub that I have a fix (I think)

AndKe:
I don't understand why my PR was not good enough ? It works, have no drawbacks, and few other improvements.

billbonney:
This is the fix.

https://github.com/billbonney/apm_planner/blob/7df90253196d17136326d39c0c7ccd95a08022fc/src/uas/ArduPilotMegaMAV.cc#L217

It translates the values depending on if you are using the severity code values or the new standardized ones.

billbonney:
PS: I will pull the other parts of the request in

billbonney:
Cherry-picked with 1b62bf5b695c6df9d114f3e00988556fcc4311c4 :+1:

billbonney:
I did some quick tests, the only issue left is that you need to restart AP2 when the option to be pre3.4 is selected. It should be an easy fix.

PR #907, opened 2016-02-20
Graph View: Fix for crash while replaying tlog - Fixes #661

### LINKED ISSUE #661: Crash during tlog playback.
Once the mission is loaded , APM planner will crash.
to reproduce: download https://drive.google.com/a/aranica.com/file/d/0B3E4K1nvpDuFVkMtUHc1YlF3WUE/view?usp=sharing

And playback at 10x speed.

Arne-W:
Made a fix for #661. The example log does not crash any more.


Arne-W:
This was not a good fix. Bad workaround and slightly buggy....

PR #914, opened 2016-03-08
Build fix for C++03

Arne-W:
Fixes: a static data member with an in-class initializer must have non-volatile const integral type - type is 'const double'

I added a prefix too - thought a constant should have one.


billbonney:
Fixed: thx :+1:

dcarpy:
Wow, that was fast.  Thanks!

Arne-W:
You are always welcome - was sitting on the sofa watching TV when I got the notification - so making a quick fix was no problem :-)

PR #915, opened 2016-03-09
Graph View: BugFix for "Graph Scale Control" issue. Fixes #859

### LINKED ISSUE #859: Graph: "Graph Scale Control" sets wrong min-max columns for new items
to reproduce:
-open a .log/.bin
-graph any log item.
-click "Graph scale control" (observe that the graphed) item have proper values in max/min columns.)
-graph another log item, (preferably with higher/lower values)
-observe that the new item is listed in "Graph Scale Control" with min/max value of the first list item.

Arne-W:
Made a fix for #859 - was a 'low hanging fruit' :-) and some refactoring.


Arne-W:
Added a basic fix for #916. The message leading to this crash is treated as error.

The crash was caused by a missing value description and a missing sanity check resulting in an out of bound index access.
At the moment I dont know how to handle this message - filter out or handle in a special way?!

Issue #919, opened 2016-03-17
Graph: "Log to google earth file" does not support .bin 


AndKe:
And while subject says all, maybe, just maybe an loaded log/bin graph is the more logical place to export from, (exporting directly from memory) - instead of reading .log and converting some of the data.

Arne-W:
Yes you are right - fixed it.

Issue #918, opened 2016-03-17
Graph: "use time on X-Axis" is not an option with Arduplane logs


AndKe:
This is related to https://github.com/diydrones/apm_planner/commit/f9fc68b5a64665bdb561692209122a3782b27c4f

Arne-W:
The "use time on X-Axis" feature is only available if every row of the log file has a timestamp. This is only supported by dataflash (.bin) logs. It should not make any difference if its a plane or a copter. 
Just send me a log (upload or mail) and I will check it.

AndKe:
here is one plane.BIN http://flyfoto-nord.no/tmp/141.BIN

Arne-W:
There are 2 problems with this log. The first one (the smaller one) is that the time colum is called "TimeMS" but the algorithm searches for "TimeUS". The second (and bigger) problem is that not all rows have a timestamp. In the Coper logs there are always timestamps which makes the handling straight forward. Creating several exeptions for with different errorhandling for planes and copters is not a good solution I think.
Perhaps the ArduPlane Guys could introduce a timestamp in every row??

AndKe:
Thank for looking into it.
yes, there is no local TimeMS for GPS and CAM messages (kind of logical, as
it has GPS time) , but MSG and PM does not have timestamp.
I asked this question in gitter , I guess that with the frequency of GPS
messages being logged, if local ms were added, that alone would grow logs
even more.

As for solution: could you just ignore these messages for x-axis time, just
take next log item ?

billbonney:
Interesting every tlog entry is time stamped on receiving in GCS and saved. 

For non time stamped messages in a bin we could just give then the same time as the last message or add 1 to that time. 

It would be good to find a solution

AndKe:
newer Arduplane does add local timestamp to GPS and CAM messages, AFAIK no change for MSG/PM , so handling of missing timestamp would help anyway.

Arne-W:
Okay - I will try to find a good solution for this... but it might take a little as the generation of artificial is not the only thing also the format description has to be patched to get a consistent datamodel. Furthermore the copter timestamps are in 'us' while the plane timestamps are in 'ms' which affects the scaling and last but not least all message types ('MODE','ERR'...) have to handle different names for the timestamp column....
I will do my best :-)

Arne-W:
Made a solution for this :-)

AndKe:
neat, thank you.

Arne-W:
[...]

PR #921, opened 2016-03-23
Time on X-Axis for all and better Google Earth (KML) export

Arne-W:
Hi Guys,

here is my solution for #918. Was not simple and not too complicated. The only problem I see is that the parser code gets more complex with every feature and maybe some day its complexity will be difficult to handle - so we should keep an eye on that.

The solution for #919 introduced a new button which makes it possible to make a direct export to a .kml file without generation a .log file first.

Take a look and let me know if you find any issues or improvements.

### DIFF
--- a/src/uas/ArduPilotMegaMAV.cc
+++ b/src/uas/ArduPilotMegaMAV.cc
@@ -394,24 +394,22 @@ void ArduPilotMegaMAV::playArmStateChangedAudioMessage(bool armedState)
 
 //******************* Classes for Sepcial message handling **************************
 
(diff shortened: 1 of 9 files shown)


Arne-W:
Added a bugfix - it could happen that a wrong mav type was detected when loading a tlog.

billbonney:
rebased onto master. Thanks :+1:

billbonney:
PS: Note we have moved locations in github to https://github.com/ArduPilot/apm_planner/

Old urls are automatically redirected, so git remotes you have setup still work, but just as a FYI.

PR #938, opened 2016-04-23
QuickView: Use HSV Color Model Fixes #932

### LINKED ISSUE #932: Widget: Info View (Quick) poor choice of colors.
Maybe something else than random would be nice?

screenshot from 2016-04-18 09-29-26

Ultimately, it would be nice to have the ability to set colors, and/or set colors based on value.

dcarpy:
HSV represents color in a way more suited to the human perception of
color. Still uses random colors, however, should work better on both
light and black background.


billbonney:
Nice. I'll check this

Arne-W:
If this gives better colours it should be used in AP2DataPlot2D.cpp - for example [here](AP2DataPlot2D.cpp#L1081) - too.

billbonney:
cherry-picked, thx

Issue #941, opened 2016-04-30
QLOG causes segfault


AndKe:
Please see attached .zip (some .tlog from forums)  (reproduced both on current master)

[20160217151228.log.zip](https://github.com/ArduPilot/apm_planner/files/243611/20160217151228.log.zip)

it ends like:
```
ERROR 2016-04-30T15:11:27.471 Corrupt data read: Time is not increasing! Last valid time stamp: "622764"  actual read time stamp is: "553687" 
ERROR 2016-04-30T15:11:27.472 Corrupt data read: Time is not increasing! Last valid time stamp: "622764"  actual read time stamp is: "553687" 
ERROR 2016-04-30T15:11:27.472 "Unable to exec getMinTime query: No query Unable to fetch row" 
 INFO 2016-04-30T15:11:27.472 Plot Log loading took 0.517 seconds - 1554368 of 1554368 bytes used 
[Thread 0x7fff8affd700 (LWP 20609) exited]

Thread 1 "apmplanner2" received signal SIGSEGV, Segmentation fault.

```

Arne-W:
Ahh - perfect already done :+1: 
I will check that too.

Arne-W:
Strange - I can't reproduce this as well... I tried the master, our crashfix branch and the new pull request - always without any issue.

AndKe:
So sorry, wrong file - I need to delete all logs in Download, I do look at a lot, then mix them up when names are similar,  - it should stop loading at ~20%   (always in same spot)
[20160318172628.tlog-tested.zip](https://github.com/ArduPilot/apm_planner/files/244019/20160318172628.tlog-tested.zip)

strangly, it varies where and how it crashes, and so resembles a bit the previous issue fixed yesterday.
this is tested on your build with the fix.

```
ERROR 2016-05-01T10:15:01.489 Corrupt data read: Time is not increasing! Last valid time stamp: "568881"  actual read time stamp is: "568815" 
ERROR 2016-05-01T10:15:01.490 Corrupt data read: Time is not increasing! Last valid time stamp: "569362"  actual read time stamp is: "569217" 
ERROR 2016-05-01T10:15:01.493 Corrupt data read: Time is not increasing! Last valid time stamp: "570806"  actual read time stamp is: "570801" 

Thread 1 "apmplanner2" received signal SIGSEGV, Segmentation fault.
0x00007ffff5b77a68 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Gui.so.5
[...]

PR #944, opened 2016-05-14
Valgrind fixes

Arne-W:
As I mentioned in #941 I have done several valgrind runs. These are the fixes for several issues valgrind complained about.
Perhaps it makes APM Planner a bit more stable but those fixes won't help for #941 I think. Anyhow I somebody likes to test - feel free :smiley:


Arne-W:
Will open a new one...

PR #950, opened 2016-05-23
QsLog and valgrind

Arne-W:
Here is my new PR. Picked the important changes from old PR #944 and added the redefinition of the QsLog macros. This should workaround the QsLog issue #941.
The other commits fix some valgrind issues, a memleak and a race condition.

### DIFF
--- a/src/ui/AP2DataPlot2D.h
+++ b/src/ui/AP2DataPlot2D.h
@@ -257,7 +257,7 @@ private:
     qint64 m_startIndex; //epoch msecs since graphing started
     QAction *m_addGraphAction;
     UASInterface *m_uas;
-    QProgressDialog *m_progressDialog;
+    QSharedPointer<QProgressDialog> m_progressDialog;
     AP2DataPlotAxisDialog *m_axisGroupingDialog;
     //qint64 m_timeDiff;
     bool m_tlogReplayEnabled;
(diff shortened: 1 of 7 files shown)


src/ui/AP2DataPlot2D.h
```diff
@@ -257,7 +257,7 @@ private slots:
     qint64 m_startIndex; //epoch msecs since graphing started
     QAction *m_addGraphAction;
     UASInterface *m_uas;
-    QSharedPointer<QProgressDialog> m_progressDialog;
```
billbonney: http://doc.qt.io/qt-5/qsharedpointer.html

QSahred pointer is always constructed as NULL and it deletes what is referenced when it goes out of scope. So the valugrind messages are not valid in this case.

Arne-W:
The problem was not the a valgrind message about a memleak. The problem was the deletion in the threadDone method while not all emitted signals were processed. So the call `loadProgress` tried to access a deleted object. The progress dialogue is now deleted by the `deleteLater` method.

billbonney:
@Arne-W Roger, that. The fix was to add deleteLater() No need to removed the QSharedPointer declaration or init it to 0. It's a shame that AP2 doesn't use more smart pointers. :)

Arne-W:
I did not want create an issue because of the deleteLater and the delete which would be done by the smartpointer, as it could not guaranteed which comes first.
But you are right, it is possible to pass a deleter method to the pointer - cool feature - I will revert the smartpointer remove today.
And you are right - too much "old school" pointers. We will change them in the future :smiley:

billbonney:
Merged with 4d0fc86f8e62c1923ec8897e4692f34070622d4e and a few previous commits. Thx :)

Arne-W:
OK - I'll close it

PR #952, opened 2016-05-29
Logging: Added messagehandler to reenable file logging.

Arne-W:
I thought we should have some sort of file logging again. 

 * one log file per day is created. Subsequent starts will be appended.
 * every appending adds a seperator for better orientation.
 * only the 5 recent logfiles will be kept, older ones will be deleted.


billbonney:
Thx :)

PR #974, opened 2016-08-12
Enlarge m_buttonStates array to prevent crash

aclex:
Enlarge m_buttonStates array member to next "maximum" value to prevent segfault in case of connecting joystick with more, than 11 buttons (reproducible e.g. with Sony DualShock 3).

Generally, if it's possible, maybe some dynamic container would be even safer here.

### DIFF
--- a/src/ui/JoystickWidget.h
+++ b/src/ui/JoystickWidget.h
@@ -100,7 +100,7 @@ private:
     Ui::JoystickWidget *m_ui;
     JoystickInput* joystick;  ///< Reference to the joystick
 
-    int m_buttonStates[11];
+    int m_buttonStates[64];
 
     QString m_buttonPressedMessage;
 


billbonney:
probably better to check here for adding greater than array size
https://github.com/aclex/apm_planner/blob/c0a305a456720a189e2d8c055337374bef6ec349/src/ui/JoystickWidget.cc#L79

aclex:
@billbonney yes, but it would lead to exclusion of several additional buttons, which otherwise could be fully working — there's even no interface artifacts in the form.

billbonney:
@aclex I mean that as an addition to increasing the size array to avoid a segfault in the future for my under 96keyed joystick ;-)

aclex:
@billbonney ahh, ok :-) surely

aclex:
@billbonney maybe, indeed should I make it more safe with QVector-ish auto-enlarged array? Just wanted keep changes as little, as possible, but generally I think, I could fix it to have no hard limit.

billbonney:
Thanks. Apologies for not getting this merged into the main code until now.

aclex:
No problem) Thanks for merging it!

PR #981, opened 2016-09-11
BUGFIX: Datamodel now accepts FMT inserts after data inserts. Fixes #980

### LINKED ISSUE #980: Graph: Error execing index: TEC2 UNIQUE constraint failed: INDEX.idx Unable to fetch row
Please see attached log.
I were asked to analyze it, (crash) - but instead discovered that APM Planner returns an error.
[12.BIN.zip](https://github.com/ArduPilot/apm_planner/files/463790/12.BIN.zip)

Arne-W:
This should fix it. It is a hack - but it works. The real problem is that the data model counts the index for the FMT messages and the parser counts the index for the data messages. Perhaps we should remove the index handling from the parser and make the index handling in the data model. I think the index is somehow an internal of the data model, so the parser should not know about.
As I'm doing a complete re factoring of the parsers at the moment I will see if I can fix it during this task.

### DIFF
--- a/src/ui/AP2DataPlot2DModel.cc
+++ b/src/ui/AP2DataPlot2DModel.cc
@@ -404,6 +404,10 @@ bool AP2DataPlot2DModel::addType(const QString &name, const unsigned int type, c
         QString variablenames = names.join(",");
         QString insertstring = makeInsertTableString(name,names);
 
+        // HACK: This line allows inserting a FMT message after inserting some data.
+        // without this a late FMT message would lead to an index failure.
+        m_fmtIndex = m_fmtIndex > m_lastIndex  ? m_fmtIndex : m_lastIndex + 1;
+
         //if (!query->prepare("INSERT INTO 'FMT' (idx,typeid,length,name,format,val) values (:idx,:typeid,:length,:name,:format,:val);"))
         m_fmtInsertQuery->bindValue(":idx",m_fmtIndex++);
         m_fmtInsertQuery->bindValue(":typeid",type);
--- a/src/ui/AP2DataPlot2DModel.h
+++ b/src/ui/AP2DataPlot2DModel.h
@@ -175,7 +175,7 @@ private:
     int m_rowCount;                 /// Stores the number of rows held in model.
     int m_columnCount;
     int m_currentRow;
-    int m_fmtIndex;
+    quint64 m_fmtIndex;
 
     quint64 m_firstIndex;
     quint64 m_lastIndex;


AndKe:
Are you sure about this line ?  :)
https://github.com/ArduPilot/apm_planner/pull/981/commits/059e7c2e108b86c9031665cf7cce1a7767131104#diff-80337e26c37772b6d75ef93ba367520fR409

Arne-W:
Okay it is a little bit overdone - an unsigned int should do the job...
One day has 86400 seconds. An unsigned int is max 4 294 967 296.... So if we log 24hours we could store 49 710 entries per second - seems to be enough. :smile: I will change that....

AndKe:
It works well, but the declaration in .h file is not needed, and gives:
```
In file included from src/ui/AP2DataPlotThread.h:38:0,
                 from src/ui/AP2DataPlotThread.cc:30:
src/ui/AP2DataPlot2DModel.h:179:13: error: redeclaration of ‘quint64 AP2DataPlot2DModel::m_fmtIndex’
     quint64 m_fmtIndex;
             ^
src/ui/AP2DataPlot2DModel.h:178:9: note: previous declaration ‘int AP2DataPlot2DModel::m_fmtIndex’
     int m_fmtIndex;
         ^
```
- otherwise, the fix works great.

Arne-W:
The `quint64 m_fmtIndex;` should replace the `int m_fmtIndex;` line... not just add...

Arne-W:
I just checked the change back to int but it leads to several type conversions as there are many other index types also a quint64. So I think it is better to leave it as it is.

AndKe:
yes, I used your fix without any change to .h file,  - simply no new declaration = success.

billbonney:
@AndKe @Arne-W Actually, you get a warning, which is what I would expect, and on 32bit systems it's going to assume int to be signed 32 and not signed 64 as most desktop/mobile these days.

I'd prefer the index was quint64 and remove the warning.

```
./apm_planner/src/ui/AP2DataPlot2DModel.cc:409: warning: comparison of integers of different signs: 'int' and 'quint64' (aka 'unsigned long long') [-Wsign-compare]
        m_fmtIndex = m_fmtIndex > m_lastIndex  ? m_fmtIndex : m_lastIndex + 1;
                     ~~~~~~~~~~ ^ ~~~~~~~~~~~
```

billbonney:
closed with 0cc6f92ae693bf975feb38340910c3eca6784d0e

Arne-W:
@billbonney It would have been better if you haven't picked this bloody hack.... opening tlogs does not work anymore.  So sorry I should have tested this :disappointed:. I will try to make a real fix - but today its too late. Perhaps you can revert the last commit....

billbonney:
It's reverted with a forced update to remove the commit

PR #982, opened 2016-09-13
RPM building scripts and assets updated

froller:
- prepare_rpm rewritten in bash instead of Perl
- .spec-file fixed to build on Fedora 24
- build-time patch added to avoid errors on RedHat-based distros


froller:
Outdated

PR #988, opened 2016-10-25
RadioCalibration: add new ELEVON_OUTPUT values

mothacehe:
ELEVON_OUTPUT values are now between 1 and 8.


mothacehe:
This patch is important because, if ELEVON_OUTPUT is between 5 et 8, apm_planner won't recognize the value and will write "-1" as new ELEVON_OUTPUT value.

billbonney:
Thx 👍

PR #995, opened 2016-12-04
Graph View: Refactored data-model for storing log data

Arne-W:
Hi Guys,
the next refactoring step is done. I have redesigned the data-storage and replaced the in memory sqlite-db by a simple set of normal data containers which gave another performance boost. Parsing binary logs is now round about 5 times faster.
Besides that the crappy index handling is now done internally by the data-storage and the Index in table view now starts at zero.


billbonney:
Thanks, this is so fast it awesome!

billbonney:
There seems to be an issue with the export logs and reading Tlogs as it doesn't account for the timestamp at a guess.

billbonney:
we can track the issue here https://github.com/ArduPilot/apm_planner/issues/997

PR #1003, opened 2017-01-15
Graph view improvements

Arne-W:
Hi,
I added an optimized method to the log-storage for getting values. It avoids an additional copy of the data. Moreover all log data that haves strings as values are suppressed in the Graph tree, as it is not useful to plot them. If we get more text messages they should be handled like error, mode,... messages.
Next I updated the copter log formatters to reflect the latest changes in ArduPilot.

### DIFF
--- a/src/ui/AP2DataPlot2D.cpp
+++ b/src/ui/AP2DataPlot2D.cpp
@@ -1033,37 +1033,16 @@ void AP2DataPlot2D::itemEnabled(QString name)
 {
     if (m_logLoaded)
     {
-        QString parent = name.split(".")[0];
-        QString child = name.split(".")[1];
-        bool isstr = false;
-        QList<QPair<double,QString> > strlist;
         QVector<double> xlist;
         QVector<double> ylist;
-        QVector<QPair<double, QVariant> > values = m_dataStoragePtr->getValues(parent, child, m_useTimeOnX);
-        if (values.size() == 0)
+        if (!m_dataStoragePtr->getValues(name, m_useTimeOnX, xlist, ylist))
         {
             //No values!
             m_graphCount++; //Prevent crash when it tries to disable
             ui.dataSelectionScreen->disableItem(name);
             return;
         }
-        for (QVector<QPair<double, QVariant> >::const_iterator i = values.constBegin(); i != values.constEnd(); ++i)
-        {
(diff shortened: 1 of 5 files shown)


Arne-W:
@billbonney could you please review - next PR is already on the way :smiley:

src/ui/AP2DataPlot2D.cpp
```diff
@@ -1074,7 +1053,7 @@ void AP2DataPlot2D::itemEnabled(QString name)
         }
         // use golden ratio for evenly distributed colors
         double golden_ratio_conjugate = 0.618033988749895;
-        double h = ((double)rand() / (double)(RAND_MAX));
+        double h = (static_cast<double>(rand()) / static_cast<double>(RAND_MAX));
```
billbonney: typically you only need to cast one side to a double to force floating point division. if static cast look long winded for typing double(rand())/RAND_MAX is my preferred style.

src/ui/LogdataStorage.cpp
```diff
@@ -221,15 +221,21 @@ double LogdataStorage::getTimeDivisor() const
     return m_timeDivisor;
 }
 
-QMap<QString, QStringList> LogdataStorage::getFmtValues() const
+QMap<QString, QStringList> LogdataStorage::getFmtValues(const bool filterStringValues) const
```
billbonney: you are passing a copy of fundamental type, it doesn't need to be declared const

billbonney:
i made some comments, which you can fix and then submit a new PR. I was kinda of playing with the new GitHub review tools..

And sorry for the delay in merging, I am just uber busy at the moment

src/ui/AP2DataPlot2D.cpp
Arne-W: casting only one side seems to be enough but using double(vale) cast leads to a clang warning about "old style cast". So I would like to stay with the static_cast<>() style.

src/ui/LogdataStorage.cpp
Arne-W: You're right - I will fix it

src/ui/AP2DataPlot2D.cpp
billbonney: interesting... c++ style casts are god for me 👍

Issue #917, opened 2016-03-15
Radio config: does not exit command mode.


AndKe:
To reproduce: 
connect a telemetry radio, and read settings more than once, you will see that it fail every-second time.
it's also prossible to leave the radio in command mode, so telemetry link won't be successful.

The radio is correctly switched into command mode using "+++"  - but then the same "+++" appears to be tried as exit.
Correct command mode exit is "ATO"

Arne-W:
The problem (one of the problems) is the the retry counter - I will fix it in the next days.
Stay tuned...

Arne-W:
@AndKe It would be great if you could test my fixes with you radio. For mine it works perfectly now.

AndKe:
@Arne-W  It works great. tested with local & remote radio's data.
good job. 

one thing remain...:
(You need only to connect one radio to reproduce)
After radio settings are read, if you unplug it, segfault follows:

```
[20170202 23:05:13.362 INFO ] - local Radio Read Complete success: true
[20170202 23:05:13.369 DEBUG] - readLocalRssi
[20170202 23:05:13.378 DEBUG] - Radio readData currentline: "ATI7\r\n"
[20170202 23:05:13.393 DEBUG] - Radio readData currentline: "L/R RSSI: 55/0  L/R noise: 77/0 pkts: 0  txe=0 rxe=0 stx=0 srx=0 ecc=0/0 temp=-276 dco=0\r\n"
[20170202 23:05:13.394 DEBUG] - Radio3DRSettings::readRemoteVersionString()
[20170202 23:05:13.394 DEBUG] - remote status: "Reading remote version"
[20170202 23:05:13.397 DEBUG] - Radio readData currentline: "RTI0\r\n"
[20170202 23:05:13.875 DEBUG] -  Failed to Read Remote Version
[20170202 23:05:13.876 INFO ] - remote Radio Read Complete success: false
[20170202 23:05:13.876 DEBUG] - remote status: "FAILED"
[20170202 23:05:19.526 ERROR] - Crtical Error! "Resource temporarily unavailable"
Segmentation fault (core dumped)

 ```

**Comment by Arne-W** [comment]
Cool - Thank you for the fast testing :+1: 
I will check that - seems to be simple to reproduce.

**Comment by AndKe** [comment]
a little more data: 

```
[20170202 23:16:38.974 ERROR] - Crtical Error! "Resource temporarily unavailable"

[...]

PR #1004, opened 2017-01-28
GraphView: Refactored log exporting / 3DR Radio Config: Bugfixes

Arne-W:
A new PR is born...

- Review fixes
- Renamed the LogParser directory to Loghandling and moved all log related files into that folder.
- Refactored the log exporting.


Arne-W:
I just added a new commit to this PR - some fixes for 3DR Radio Config.
Here is a list:
* link will be only set if changed
* added sleep between the remote and local reset (otherwise the remote reset won't be transmitted)
* corrected setup of remote frequency boxes (could be discussed whether it is useful to set different frequencys on remote than on the local radio)
* filter unwanted ascii data
* disconnect from signals before reconnect
* better detection of reset
* correced handling of disconnect (won't crash anymore)
* leave commandmode on close of connection; reset retrycounter on disconnect.

Thats it :smiley:

Arne-W:
Ok now all pointer access has a sanity check - should not crash any more. I also added a message box if the port gets lost.

Arne-W:
Just for info - see #917 for discussion about the 3DR Radio Config.

Arne-W:
Looks like its all done now - could be reviewed and merged....

Issue #1005, opened 2017-02-06
Graph: Ideas for even better log analysis.


AndKe:
AP2 is (by far) the greatest log analysis tool I know. It's speed, smooth navigation is great.

please bear with me for a feature request that would make it even better:
During tuning , writing of POH, gathering performance data or general log analysis, we often need to know an average, or difference during a certain maneuver or time.   

Please see mockup. 
If a checkbutton is selected, two red lines appear, they can be dragged to select area of interest.
the mouseover displays the time between those lines, and we can sea delta (difference) for value during that time frame, and it's average.

Usage examples:
Delta:
- would easily show acceleration (accX) range during normal flight, before some failure that introduced extra vibrations. (and also easily qualtify those , even if the whole graph remain scaled to some other event) 
- find climb rate/descend rate, roll or pitch rate  - when something is happening, let's say roll rate in a turn, or descent rate. 
-find power consumption over xx seconds at xx speed.

Average: 
-easily quantify how much an axis is changing during apparently "flat" logline,  select 10s , see how much variation we have in pitch.
-get proper average throut , used for documenting multirotors,  1kg payload = throut xx  , 2kg=throut xx  - then we know it's maximum capacity vs flight times m vs redundancy reserves. 
-Also, during very smooth hover in still air, properly tuned multi has low variations in throut,  - all rcout.chX should have similar average (and delta)   - otherwise it's a indication for some problem. 
 
The tools, as illustrated, will aid proper log analysis a lot.

![screenshot from 2017-02-06 22-17-41](https://cloud.githubusercontent.com/assets/1468842/22668322/9ddae066-ecc0-11e6-8840-c7139270419a.png)

Arne-W:
Cool idea - should be done! :+1:

Arne-W:
Started to work on this feature, but the classes which are responsible for the graph view need some refactoring cause they get bigger and bigger.
So far I have updated to a new QCustomPlot version and split the the Graph view code in 2 classes - one for online viewing and one for offline log analysis. Next thing will be the cursor...
I thought about a context menu when clicking into the graph to add one or two cursors.
I will tell you when the feature is testable...

AndKe:
Great :)

Arne-W:
Ok its nearly done :smile: If somebody has some time to waste he could check [THIS](https://github.com/Arne-W/apm_planner/tree/New_LogAnalysisWindow) branch.
New Features:
* Range cursors can be used simultaneously with old cursor.
[...]

Issue #646, opened 2015-03-18
CPU hogging after connect, 


AndKe:
APM planner takes about 30% CPU time on a quad core 2.4Ghz laptop ftom the moment of serial connection.
Disconnecting makes it drop less then 5% , to about 26%
This goes also for non-graphic views whithout maps/AHI

malcom2073:
Relatively high cpu usage is expected, but it should not be continuing after disconnection. I am able to reproduce this on Windows, so I'll try to figure out what's spinning while disconnected.

AndKe:
Confimed by another user here: https://github.com/diydrones/apm_planner/issues/716

AndKe:
how did this issue get cloased ?
It's still hogging CPU like crazy, and extended laptop battery can't help that.
Starts after telemetry connection.

AndKe:
To reproduce:
-connect to pixhawk using USB or radio.
-observe the *application* occupy for example 15% CPU within 1 munite of connection
-disconnect , remove USB radio/Pixhawk , observe that CPU use continues (by apmplanner2)to climb
- vait 10 minutes, , you may easily be at 35% cpu by then...

please reopen..

AndKe:
Additional info: "Primary flight display" is most likely to blame.
Reproducable:
If I allow the CPU hogging  problems to start, closing "primary flight display" widget does not help. (maybe a clue here!?)
If I close "primary flight display" before connecting/changing tabs, then the problem will not appear.  
Also: "primary flligt display (old)" does not cause this.
Finally: just using "primary flight display" does not cause the problem by itself, changing tabs and then back to flight data tab where it is does.

AndKe:
Final additional info:
The GCS laptop is a Latitude 5520 , i7 CPU , Ubuntu 15.04 x64 , with intel graphics(CPU)
"normally" working APM Planner2 is consuming about 3% CPU, a little more than playing a HD video.

my workstation has a faster CPU, and Nvidia GPU - usual apmplanner2 CPU usage is <1% , and the problem is *not* reproducible on this system.

So we could blame it on the Intel graphics, except I have not found any  ohter app behaving this way, and it's caused by this one widget. 

Also, tab switching is trigging the problem, but also minimizing and restoring the apmplanner2 window.

billbonney:
I'll see if I can reproduce this on my linux machine.

Q: If you reconnect to the serial device, to another serial device, does the CPU usage return to normal?
Q: Have you tried updating the graphics drivers
[...]

PR #1017, opened 2017-03-20
GraphView: Added range cursor feature and did some refactoring

Arne-W:
Hi all,
here is the PR for the Range Cursor feature requested in #1005.
Moreover this PR includes:
 * All cursor moving can be done by drag and drop. Selecting a cursor is done by a click
 * Cursors can be enabled by checkbox or context menu (right click into graph area).
 * Value axis can be scaled with mouse wheel, must be clicked first to select and mouse must be over axis while using mouse wheel.
 * The range cursors deliver min, max, delta and average for all enabled graphs.
 * Splitted "AP2DataPlot2D" into two classes: AP2DataPlot2D - for the live log view and LogAnalysis for all offline analyzing. 
 * Migrated to QCustomPlot 2.0.0-beta (will update to final as soon as it is released)
 * Renamed some features, as their names where misleading.

Here is a screenshot:
graphview

### DIFF
--- a/apm_planner.pro
+++ b/apm_planner.pro
@@ -448,7 +448,8 @@ FORMS += \
     src/ui/configuration/CompassMotorCalibrationDialog.ui \
     src/ui/MissionElevationDisplay.ui \
     src/ui/uas/PreFlightCalibrationDialog.ui \
-    src/ui/configuration/RadioFlashWizard.ui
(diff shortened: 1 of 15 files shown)


Arne-W:
Added the partial fix for #646. Second part (start script for linux) has still to be done. This change should not have any impact if used without the script.

Issue #1001, opened 2017-01-12
changes to copter binary names


tridge:
Hi Bill,
This is a heads up that binary names for copter builds are changing, which will affect firmware upload
I've created a MP issue describing it here:
https://github.com/ArduPilot/MissionPlanner/issues/1456
Cheers, Tridge
CC: @rmackay9

mboland:
Mission Planner can no longer find the binaries for 3.5 beta firmware.
[Forum Issue](http://discuss.ardupilot.org/t/cannot-get-beta-copter-to-appear-in-firmware/16051)

Just wondering when might expect a fix for this?

billbonney:
@Arne-W DO you think you can look at this, sorry I don't have time at the moment

Arne-W:
@billbonney Yes no problem - I will check that tomorrow. BTW you should take a look at #1000 I made a WIP - if yo like it I will make a PR.

Issue #1000, opened 2016-12-31
Support new FRAME_CLASS, FRAME_TYPE parameters for Copter-3.5


rmackay9:
For AC3.5 we are introducing a FRAME_CLASS parameter and renaming FRAME to FRAME_TYPE so it would be good to update the frame selection screen.

Possible values from FRAME_CLASS are 0:Undefined, 1:Quad, 2:Hexa, 3:Octa, 4:OctaQuad, 5:Y6, 6:Heli, 7:Tri, 8:SingleCopter, 9:CoaxCopter
Possible values for FRAME_TYPE are 0:Plus, 1:X, 2:V, 3:H, 4:V-Tail, 5:A-Tail, 10:Y6B

Here's the equivalent to-do for MP which includes a mock-up of what the screen might look like: https://github.com/ArduPilot/MissionPlanner/issues/1442

Arne-W:
@billbonney shouldn't we start implementing this feature? I saw your [comment](https://github.com/ArduPilot/MissionPlanner/issues/1442#issuecomment-269847367) but it seems that nobody tries to find a better solution.
Perhaps we should show only the valid FrameTypes for a selected FrameClass. This would make it impossible to select undefined combinations. The drawback is that we will have to adapt the selections whenever ArduPilot adds or removes some combinations.
I started some implementation tests (very early stage)- see [https://github.com/Arne-W/apm_planner/tree/NewFrameType_WIP](https://github.com/Arne-W/apm_planner/tree/NewFrameType_WIP). At the moment it gives you a blank frame screen when ArduPilot 3.5 is detected. Next steps will be to fill the screen.

Arne-W:
@rmackay9 Which FRAME_TYPE has to be selected for Heli and Tri or do they not use the TYPE parameter?

Arne-W:
All right - did not take as much time as I thought. If anybody has some time it would be nice if he could test the Branch mentioned above. There is still some clean up to do but the functionality should be complete. The branch should work with all ArduPilot FW versions.

rmackay9:
@Arne-W, thanks very much for doing this.

I had a peek at the Copter motors library and Heli and Tri copters both ignore the FRAME_TYPE.
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Motors/AP_MotorsTri.cpp#L29
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Motors/AP_MotorsHeli.cpp#L182
[...]

PR #1021, opened 2017-03-29
Support ArduCopter Firmware 3.5.0

Arne-W:
This PR adds support for ArduCopter 3.5.0
 * Supports the new frame configuration parameters "FRAME_CLASS" and "FRAME_TYPE". #1000
 * Support for the changed filenames in ArduCopter. #1001 

I have tested frame configuration against SITL and the filename adaption till a successful download. Could not test the flashing itself as I don't have a test target at the moment.


Arne-W:
I have added the single and coax copter to the frame selection. Should be complete by now, only the icons for Single / Coax Copter, V-Tail and A-Tail are missing. Perhaps someone can create some or knows where I could get them.

Issue #1024, opened 2017-04-02
Plan/Flight data , orphaned lines between waypoints.


AndKe:
This can only be reproduced when receiving mission from AP , not by loading mission file.

To reproduce: 
1: Start AP2 ,Connect to autopilot that has a mission(or SITL). move the map a little, 
2: The mission will automatically show up in "Flight data" 
3: move the map a little, in case of this mission , observe the two orphaned red lines south-east of the mission , orphans/clones of lines between takeoff,WP1,WP2

![screenshot from 2017-04-02 10-54-52](https://cloud.githubusercontent.com/assets/1468842/24588028/faf80fe6-17c0-11e7-9c2d-d2ad86075b0a.png)

4: switch to flight plan.
5:move map a little, observe that there is another set of "red lines" stuck

The orphaned red lines are stuck to the "window",  panning or zooming does not affect them.
![screenshot from 2017-04-02 10-55-23](https://cloud.githubusercontent.com/assets/1468842/24588102/4d6f206a-17c2-11e7-9909-ad2b675ac194.png)

Also attaching the mission in this test, but it seems to happen to very mission. 
[snarbytest.txt](https://github.com/ArduPilot/apm_planner/files/888417/snarbytest.txt)

Issue #1023, opened 2017-04-01
Graph: where did "filter" go ?


AndKe:
I miss the "Filter" feature that filtered the datagrid lines by MSG type.
Was it removed intentionally, or by accident when adding new graphing features  ?

Arne-W:
Which "Filter" do you mean?
![filter](https://cloud.githubusercontent.com/assets/7621911/24582876/28c78e60-173a-11e7-95d1-48124c2f3c3c.jpeg)
This one?

AndKe:
Yes , do you see it in current master ? (I don't)

Arne-W:
As the current master has currently changed you should try again - and yes - I can see it.

AndKe:
Very strange, again, I just built clean from master, (2.0.20-149-g5334dbe)

here's how it looks - and no, increasing width does not reveal the button I miss. 
![screenshot from 2017-04-02 16-10-11](https://cloud.githubusercontent.com/assets/1468842/24587926/4fc71dfc-17bf-11e7-924c-9905d0b03182.png)


what I do is open apmplanner2 ,click "graphs", "open log", and load a  .BIN
I've also deleted the ini file.

Arne-W:
Hmm strange - when I resize the graph window to the smallest possible size it looks like this:

![filter1](https://cloud.githubusercontent.com/assets/7621911/24589730/c5d183c8-17df-11e7-9ff6-a1b193c0e42c.jpeg)

I can not make it smaller.
I think I can make a test version with resized buttons - I will inform you!

AndKe:
Please note that it's not about "small window",   I have normal size window
just the same, no amount of resizing (or fullscreen window) reveal that
button.

mboland:
Is this in Daily build yet?
Or do I need to compile Master?
I have 3 different screen sizes from largest to small I can test on.

Arne-W:
Yeah - it should be. Would be nice if you could have a look.

mboland:
No Filter button visible.
Details:
AMD FirePro D700:
  Chipset Model:	AMD FirePro D700
  Type:	GPU
  Bus:	PCIe
  Slot:	Slot-2
  PCIe Lane Width:	x16
  VRAM (Total):	6144 MB
  Vendor:	ATI (0x1002)
  Device ID:	0x6798
  Revision ID:	0x0000
  ROM Revision:	113-C3861J-687
  Automatic Graphics Switching:	Supported
[...]

Issue #1022, opened 2017-04-01
Graph: .BIN Logs made with GPXS , GRXH, EKF5 causing slow performance and lots of errors.


AndKe:
[GPS.zip](https://github.com/ArduPilot/apm_planner/files/887828/GPS.zip)

please see attached file, the full file ~80MB takes two minutes to load, and gives >500k "corruption" warnings.

Arne-W:
Okay - one of the problems is that I forgot to implement the parsing of double values :grimacing:  I made a fix for that...
The second problem is that several float and double values resolve to "Not A Number" which results in a drop of the message.
And last but not least - the performance drop is a result of the massive logging in case of such systematic issues.

Issue #1019, opened 2017-03-24
FlightPlan:multiple waypoints are moving together after load.


AndKe:
To reproduce; 
-load (from file) or read (over mavlink) a mission
-try to move, as in fine-adjust individual mission items.
- observe that more than one is moving
- after moving enough of them, or depending on which you move, it will start behaving as expected.
- issue returns on mission read or load.

see video : https://youtu.be/O5MuNFnxrG0

PR #1025, opened 2017-04-04
Several Fixes

Arne-W:
A new PR is born :smiley: 
I have made several fixes for the following issues:
 * #1019 - FlightPlan:multiple waypoints are moving together after load.
 * #1022 - Graph: .BIN Logs made with GPXS , GRXH, EKF5 causing slow performance and lots of errors. 
 * #1023 - Graph: where did "filter" go ? 
 * #1024 - Plan/Flight data , orphaned lines between waypoints.


Arne-W:
Added another commit to allow better diagnosis on way point reception and transmission.

Arne-W:
As there is a  bug in this PR I will close it and create a new when Bug is fixed

PR #1028, opened 2017-04-07
Several Bug fixes

Arne-W:
After fixing the bug made in last PR here is the replacement for #1025 

Here is the list of what have been fixed:
 * #1019 - FlightPlan:multiple waypoints are moving together after load.
 * #1022 - Graph: .BIN Logs made with GPXS , GRXH, EKF5 causing slow performance and lots of errors. 
 * #1023 - Graph: where did "filter" go ? 
 * #1024 - Plan/Flight data , orphaned lines between waypoints.

I hope this time nobody finds a bug :smiley:

### DIFF
--- a/libs/opmapcontrol/src/internals/core.cpp
+++ b/libs/opmapcontrol/src/internals/core.cpp
@@ -173,6 +173,7 @@ namespace internals {
                                         QMutex m;
                                         m.lock();
                                         wait.wait(&m,500);
+                                        m.unlock();
                                     }
                                 }
                             }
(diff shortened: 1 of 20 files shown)


Arne-W:
Added another commit which fixes a bug when grouping graphs. Some min max values were not correctly initialized which leads to wrong scaling.

Arne-W:
Here we go 2 new commits.
First one fixes the ugly c++11 warning and the second improves the readability of status text on waypoint transfer.

Arne-W:
another commit - fixing an issue introduced with orphaned lines fix

PR #1039, opened 2017-04-26
WIP: add feature: plane timestamps

kd0aij:
@Arne-W This is a small mod to your bugfix branch to add GPS time to plane placemarks.
I had to robustify the GPS record validation check because I was getting records with zero week and week_msec fields, and that screws up Google Earth.
If it's working correctly, you should get a "time slider" at the upper left of the GE map display as soon as you load a KMZ file.

This could easily be added to the flight path data also.

### DIFF
--- a/src/output/kmlcreator.cc
+++ b/src/output/kmlcreator.cc
@@ -184,7 +184,9 @@ QString SummaryData::summarize() {
 }
 
 KMLCreator::KMLCreator():
-    m_summary(new SummaryData()) {
+    m_summary(new SummaryData()),
+    m_newGPSMessage(false)
+{
 }
 
 KMLCreator::~KMLCreator() {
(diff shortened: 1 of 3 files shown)


src/output/kmlcreator.cc
```diff
@@ -491,7 +491,32 @@ void KMLCreator::writePlanePlacemarkElement(QXmlStreamWriter &writer, Placemark
 
     int index = 0;
     foreach(GPSRecord c, p->mPoints) {
+        std::string week_ms_str = c.msec().toStdString();
+        const char *tschar = week_ms_str.c_str();
+        char* endptr;
+        uint32_t week_ms = strtoul(tschar, &endptr, 10);
```
Arne-W: fetching the week_ms can be done much easier: `int weekMS = c.msec().toInt();` for example.

src/output/kmlcreator.cc
```diff
@@ -491,7 +491,32 @@ void KMLCreator::writePlanePlacemarkElement(QXmlStreamWriter &writer, Placemark
...
+
+        bool status;
```
Arne-W: the return value is not processed - remove it please

kd0aij:
That code is a quick & dirty hack just to see if the timestamps would work as expected, and I assumed week_ms was unsigned for a good reason.
I also assume there's a utility around somewhere to do a proper conversion from GPS iTOW to UTC time, without the need for reinventing one. Need to find that next.

Arne-W:
Ah - all right :+1:

PR #1040, opened 2017-04-27
WIP: enhancement, add timestamps to flightpath and plane placemarks

kd0aij:
supersedes #1039
rebased on master

cleaned up the GPS to UTC time conversion, now matches time_epoch_convert() in  AP_GPS.h

Also breaks up flight path into timestamped 1 second segments

### DIFF
--- a/src/output/kmlcreator.cc
+++ b/src/output/kmlcreator.cc
@@ -10,6 +10,7 @@
 #include <QProcess>
 #include <QWaitCondition>
 #include <QMutex>
+#include <QQuaternion>
 #include <quazip.h>
 #include <math.h>
 
@@ -95,6 +96,7 @@ static QString toModeString(QString &line) {
     QStringList parts = line.split(QRegExp(","), QString::KeepEmptyParts);
 
     // TODO: Need to add Mode: DataLine object to fix
+    // ... create a ModeRecord:DataLine class and take vehicle type into account
     // MODE, 82081720, 5, 5, 1 // New Message with time
     // MODE, ALT_HOLD, 516     // Old message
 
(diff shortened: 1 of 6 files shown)


Arne-W:
Cool - I like this animation stuff

kd0aij:
@Arne-W I wanted it to analyze a VTOL transition that didn't go well; this is the example: 
http://discuss.ardupilot.org/t/dual-motor-tailsitters/15302/312?u=kd0aij

But there's an issue with VTOLs; the Euler angles in the ATT log record are calculated from a rotated reference frame when a tailsitter is in hover mode (nose vertical is defined to be zero pitch instead of 90 degrees). Since the rotated attitude is wrong for the KML plane icon, I've added support for AHR2 (which is unrotated) and two  new records XKQ1 and NKQ1 which contain unrotated quaternions. Currently, the kml generator just assigns a fixed priority to these records, but we'll probably want to allow user selection instead.

kd0aij:
@Arne-W I reduced the size of the plane icon. Let me know if it still looks too large in your logs. We may want to allow user control of the distance between successive icons to avoid excessive overlap when velocity is low.

Arne-W:
Awesome - the kml export gets professional :+1:
The plane size is reduced - but still huge. I also tried the scaling with the scaling parameter but to scale it to your size the factor has to be smaller than 0.1 and made the model look somehow strange.
Its a nice Idea to let the user configure the export. We could extend the export menu in the loganalysis window by an "configure KML Export" button, which calls a method in the KMLexporter opening a settings window. The KML export has to store this settings in the .ini file and load them on construction.

kd0aij:
@Arne-W Thanks for the advice on implementation; I'll try adding a button soon.

Arne-W:
@kd0aij Just an Idea - if you need some help - let me know

billbonney:
@kd0aij Can you fix up the conflicts and then I will merge in the code.

if #1039 is not required to be merged can you close it? thx 👍

kd0aij:
Sorry I forgot to close #1039 earlier. This is still WIP, but I'll try to get it into a mergeable state ASAP.
More work is necessary to display the unrotated plane icons correctly near 90 degrees pitch.

kd0aij:
@Arne-W This is what I'm seeing in GE on Ubuntu 16.04 (animation of EKFattitude plane models):
https://drive.google.com/open?id=0Bw3digSMQXDuUmJnZ0ZSNFFNUUE

kd0aij:
@Arne-W  @billbonney rebased on master, but probably should review before merging. There are a few bits of code which are currently unused, but will be needed when we implement the "configure KML" button. I've changed the name of the "Planes" folder in the KML file to be more descriptive, since there are now two of them, and potentially several more will be added.

kd0aij:
@Arne-W Could you provide (or point me to) some older logs which would be good test candidates? I've only recently started using Ardupilot and don't have any myself. Did older logs not have the timeUS field in every message, ?

The new commit passes the MAV_TYPE into the KMLCreator in order to interpret the flight modes correctly. I think I implemented it in a reasonable manner, but my C++ is a bit rusty.

kd0aij:
@Arne-W Finally figured out how to get the elevation profile to work again; added a 50% transparent grey unsgemented flight path:
elevprofile

Arne-W:
@kd0aij looks good - on weekend I will search for some older logs.

Arne-W:
@kd0aij Here are some older logs. One 3.1.5, one 3.2.1 and one 3.3. The 3.1.5 and the 3.2.1 do not have a us timestamp just ms.
[logs.zip](https://github.com/ArduPilot/apm_planner/files/981583/logs.zip)

kd0aij:
@Arne-W Thanks, I'll try to make sure the KML generator behaves reasonably with the older logs. I suppose the code version is available only in the text fields of the startup MSG records?

Arne-W:
@kd0aij Yes you are right - only in the startup MSG. The main difference is the name of the timestamp if you can handle the different ones you do not have to check the version I think. Some data will be missing in the export, but if there is no data you can't do anything

kd0aij:
@Arne-W I also added a simple dialog box for specifying the min distance between EKFattitude placemarks. That's in lieu of a more elaborate configuration menu and .ini file, since ATM there's only the one parameter.

Arne-W:
@kd0aij as long as there is only one Parameter it is perfect.
I would like to see this Feature in the next release. Thanks for the work :+1:

kd0aij:
@Arne-W Thanks, glad to be able to contribute.

Arne-W:
@kd0aij Do you have still any additions to this PR? If not we should inform bill that he can merge it.

kd0aij:
@Arne-W @billbonney I think it's ready to merge. There are a few enhancements that would be nice, but I'll have to learn a lot more about KML first. And it would be nice to get some user testing and feedback before that.

billbonney:
thanks, merged 👍

PR #1048, opened 2017-06-08
Build: Fedora 25 RPM build fixes

froller:


redhat/apmplanner2-0.patch
```diff
@@ -0,0 +1,63 @@
+diff --git a/apm_planner.pro b/apm_planner.pro
```
larsks: I've submitted this as a separate pr (#1053).  I think it makes more sense to have this particular change committed to the apm planner sources, rather than carrying it as a patch in the rpm.

froller: Kinda RedHat approach. Keeping original source as is and issuing all distro-specific changes as patches. Thus I did so.

larsks:
The Red Hat approach would be "contribute changes upstream when possible " :). I'm actually working on a larger set of changes related to packaging for Fedora; I'll try to get them posted today or tomorrow so you can let me know what you think.

froller:
Do you think tearing flite off the origin is a good idea? Seems Ubuntu and other Debian guys will hardly be happy with that.
So why don't we leave this as a patch until appropriate solution?

Arne-W:
@froller could you please rebase this PR onto master? We have merged several PRs and I do not have a Fedora distri to check if your changes still work. Sorry for the inconvenience - I promise we will merge your PR as soon as it is rebased.

Arne-W:
@froller would you please check if all your changes are still useful? As I wrote above - it would be great if you could check what we should still merge - a lot of changes got in since you opened this PR.

Arne-W:
as discussed in #1135 I will close this now

PR #1052, opened 2017-06-16
fix incompatibility w/ openssl 1.1.x

larsks:
OpenSSL 1.1.0 hid structure internals and provides methods for accessing them. This patch adapts the code so that it can be built against new and old OpenSSL, which allows apm_planner to build successfully on Fedora 26.

(solution adopted from https://rt.cpan.org/Public/Bug/Display.html?id=118345)


billbonney:
Thanks 👍

PR #1053, opened 2017-06-16
build: fix apm_planner.pro for building on fedora/rhel/etc

larsks:
The version of flite available on Fedora (etc) is incompatible with
apm planner code.  This commit modifies apm_planner.pro to exclude
flite support when building on these platforms.


larsks:
I have a better solution for this issue that I will post later this week.

PR #1065, opened 2017-07-31
build cleanups and bugfixes

larsks:
- introduce a build-time test for flite support to determine if a
  compatible version of flite is installed
- remove duplicate code from apm_planner.pro
- correct spelling of redhat-release in apm_planner.pro
- make installation of debian menu files only happen on debian/ubuntu
  environments
- move sik firmware into /usr/share


larsks:
@Arne-W any chance you (or someone else?) can take a look at this?

Arne-W:
@larsks I will have a look this evening.

Arne-W:
@larsks I forgot - this PR is the "state of the art" solution for the same issue addressed in PR #1048?

larsks:
I think this is a better way of solving the issue involving `FLITE_AUDIO_ENABLED`. I think #1048 should be rebased once this patch merges, because it has a bunch of rpm build stuff in it that may still be useful.

billbonney:
Thx 👍

PR #1070, opened 2017-10-26
 Avoid downloading deprecated apm.pdef.xml 

peterbarker:
... use vehicle-specific files instead


src/ui/configuration/ApmSoftwareConfig.cc
```diff
@@ -163,7 +183,12 @@ void ApmSoftwareConfig::apmParamNetworkReplyFinished()
 
     } else {
         QByteArray apmpdef = reply->readAll();
-        m_apmPdefFilename = QDir(QGC::appDataDirectory()).filePath("apm.pdef.xml");
+
+        QDir autopilotdir(QGC::shareDirectory() + "/files/" + "ardupilotmega");
```
Arne-W: Why did you change the data path from `appDataDirectory()` to `shareDirectory()`? On my Ubuntu 16.04 `shareDirectory()` delivers the path to the binary of apm planner - on some systems the User may not have the rights to write to that folder. Moreover the path is extended by `/files/ardupilotmega` but not checked if valid, if it does not exist it will **not** be automatically created and therefore the `file.open(QIODevice::ReadWrite | QIODevice::Truncate)` will always fail.
I would prefer using the old path which is in `home/apmplanner2` and perhaps we should create a subfolder?!

src/ui/configuration/ApmSoftwareConfig.cc
```diff
@@ -290,11 +327,12 @@ void ApmSoftwareConfig::activeUASSet(UASInterface *uas)
 #else
     QString appDataDir = getenv("HOME");
 #endif
-    m_apmPdefFilename = QDir(appDataDir + "/apmplanner2").filePath("apm.pdef.xml");
+    QDir autopilotdir(QGC::shareDirectory() + "/files/" + uas->getAutopilotTypeName().toLower());
```
Arne-W: see above

Arne-W:
@peterbarker If you don't have the time I can make the changes I requested - just wanted to be sure that changing the path does not break anything on other systems...

peterbarker:
@Arne-W Sorry, I have been meaning to get back and study the changes again.

I *did* reorder things to the vehicle-specific XML files should be looked at first (falling back to an existing `apm.pdef.xml`)

I did *not* mean to change the path the the all-inclusive, all-messed-up `apm.pdef.xml`.

Apparently I did not manage both of those!

Arne-W:
@peterbarker Ahh Ok  :smile:  just let me know if I shall fix this

peterbarker:
On Mon, 6 Nov 2017, Arne Wischmann wrote:

> @peterbarker Ahh Ok 😄 just let me know if I shall fix this

Thanks Arne, if you could do that that would be great.  Kind of busy ATM.

Arne-W:
@peterbarker No problem - I will fix it

PR #1074, opened 2017-11-04
Map: added Eniro as map for Norway,Sweden,Finland,Denmark,Poland

AndKe:
Nice, detailed maps.


Arne-W:
Nice :+1: 
Thank you.

PR #1082, opened 2017-12-14
Some small fixes in opmapcontrol

hdino:
While using the opmapcontrol lib in another project I came across some issues like memory leaks and deprecated project files. The committed changes fix those.

### DIFF
--- a/libs/opmapcontrol/src/internals/core.cpp
+++ b/libs/opmapcontrol/src/internals/core.cpp
@@ -67,8 +67,8 @@ namespace internals {
     {
         ProcessLoadTaskCallback.waitForDone();
         Matrix.Clear();
-        if (projection)
-            delete projection;
+        delete projection;
+        projection = 0;
     }
 
     void Core::run()
--- a/libs/opmapcontrol/src/internals/core.cpp
+++ b/libs/opmapcontrol/src/internals/core.cpp
@@ -42,6 +42,7 @@ namespace internals {
     minOfTiles(0,0),
     maxOfTiles(0,0),
     zoom(0),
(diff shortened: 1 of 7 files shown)


libs/opmapcontrol/src/internals/core.cpp
```diff
@@ -65,6 +66,9 @@ namespace internals {
     Core::~Core()
     {
         ProcessLoadTaskCallback.waitForDone();
+        Matrix.Clear();
+        if (projection)
```
Arne-W: deleting a null-pointer is a nop. So I would prefer something like this:
`delete projection;
projection = 0;`
Another possibility would be using an autoPointer.

libs/opmapcontrol/src/internals/core.h
```diff
@@ -135,6 +135,9 @@ namespace internals {
         }
         void SetProjection(PureProjection* value)
         {
+            if (projection)
+                delete projection;
```
Arne-W: See above

libs/opmapcontrol/src/internals/pureprojection.h
```diff
@@ -44,6 +44,8 @@ class PureProjection
 
 
 public:
+    virtual ~PureProjection() {}
```
Arne-W: :+1:  very good finding!

hdino:
Thanks for your review! I didn't want to use an autopointer since the whole lib uses raw pointers. I think I would be better to change all pointers at once to keep the code consistent.

Arne-W:
Thanks for the update - and yes only switching one pointer to an automatic one could be confusing.

PR #1083, opened 2017-12-19
Flightplan: refreshPos functions of graphic items call method changed

hdino:
This change allows to add new graphical items to the map without having to modify mapgraphicitem.cpp all the time. Moreover, I did not like the try-and-error approach to cast the graphic items to all types and then check whether the cast was successful or not.


Arne-W:
Just a thought - why not introduce an interface for those items and make the refreshPos() method pure virtual? This would force everyone to implement such a function and it could be called through the interface on all derived types. Moreover it is checked at compile time if the interface is implemented and reduces the coupling.

hdino:
Sounds good to me. I am going to push a patch.

PR #1087, opened 2018-01-22
NaN handling and minor changes

Arne-W:
A new PR for handling the NaNs in flash (*.bin) logs. Besides that two minor fixes are introduced as well. One adds some output to the log file which were only shown in GUI, the other one fixes the one wrong log output which always marked a successful log loading as cancelled.


Arne-W:
commits should have prefix in comment

PR #1088, opened 2018-01-22
NaN handling and small fixes

Arne-W:
A new PR for handling the NaNs in flash (*.bin) logs. Besides that two minor fixes are introduced as well. One adds some output to the log file which were only shown in GUI, the other one fixes the one wrong log output which always marked a successful log loading as cancelled.

### DIFF
--- a/src/ui/Loghandling/BinLogParser.cpp
+++ b/src/ui/Loghandling/BinLogParser.cpp
@@ -38,8 +38,7 @@ bool BinLogParser::binDescriptor::isValid() const
     {
         if(m_format.size() != m_labels.size())
         {
-            QLOG_WARN() << "binDescriptor::valid() Corrupt FMT descriptor found - known bug in some logs - "
-                        << "trying to ignore...";
+            QLOG_WARN() << "binDescriptor::valid() Corrupt FMT descriptor found - known bug in some logs";
         }
         return (m_ID != 0xFF) && (m_length > 0) && (m_name.size() > 0) &&
                (m_format.size() > 0) && (m_labels.size() > 0);
@@ -49,8 +48,7 @@ bool BinLogParser::binDescriptor::isValid() const
     {
         if(m_format.size() == 0 && m_length == 3)
         {
-            QLOG_WARN() << "binDescriptor::valid() Corrupt STRT descriptor found - known bug in some logs - "
-                        << "trying to ignore...";
+            QLOG_WARN() << "binDescriptor::valid() Corrupt STRT descriptor found - known bug in some logs";
         }
         return (m_ID != 0xFF) && (m_length > 0) && (m_name.size() > 0) &&
                (m_format.size() == m_labels.size());
@@ -180,7 +178,7 @@ AP2DataPlotStatus BinLogParser::parse(QFile &logfile)
     }
     if (noMessageBytes > 0)
     {
-        QLOG_DEBUG() << "BinLogParser::parse(): Non packet bytes found in log file. " << noMessageBytes << " bytes filtered out. This may be a corrupt log";
+        QLOG_WARN() << "BinLogParser::parse(): Non packet bytes found in log file. " << noMessageBytes << " bytes filtered out. This may be a corrupt log";
         m_logLoadingState.setNoMessageBytes(noMessageBytes);
     }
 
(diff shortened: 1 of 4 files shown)


src/ui/Loghandling/BinLogParser.h
```diff
@@ -76,6 +76,12 @@ class BinLogParser : public LogParserBase
     static const int s_FMTFormatSize = 16;       /// Size of the format field in FMT message
     static const int s_FMTLabelsSize = 64;       /// Size of the comma delimited names field in FMT message
 
+    static const quint32 s_FloatNaNDetector     = 0x7FC00000;   /// Value to mask and detect all float NaN from ardupilot
+    static const quint32 s_FloatSoftNaNDetector = 0x7FC04152;   /// Value to detect a quiet/soft float NaN from ardupilot
```
peterbarker: These are, indeed, the only magic numbers that ArduPilot currently produces.  They are AP specific, kinda, in that one says "AP" and the other "ARDU" IIRC.

If another autopilot were to add their own soft nans then this check would need to be changed.

I like the specific check as it currently is, just mentioning this fact.

src/ui/Loghandling/BinLogParser.cpp
```diff
@@ -332,14 +349,32 @@ bool BinLogParser::parseDataByDescriptor(QList<NameValuePair> &NameValuePairList
...
-                QLOG_WARN() << "Corrupted log data found - double resolves to NAN - Graphing may not work as expected for data of type " << desc.m_name;
-                m_logLoadingState.corruptDataRead(static_cast<int>(m_MessageCounter), "Corrupt data element found when decoding " + desc.m_name + " data.");
-                NameValuePairList.clear();
-                break;
+                if (*valPtr == s_DoubleSoftNaNDetector)
+                {
+                    QLOG_DEBUG() << "Double resolves soft NaN - meaning this value contains no valid data!";
+
+                    // Store a Qt Quiet NaN in the data - should be handled correctly by the graphing toolset
+                    val = qQNaN();
```
peterbarker: Nice Qt supplies one of these.  Python made life difficult.

Arne-W:
@OXINARF Thank you for the fast response :+1:  Hopefully @peterbarker likes it too :wink:

src/ui/Loghandling/BinLogParser.cpp
```diff
@@ -332,14 +349,32 @@ bool BinLogParser::parseDataByDescriptor(QList<NameValuePair> &NameValuePairList
         }
         else if (typeCode == 'd')
         {
+            // we want to read a double -> 8byte -> DoublePrecision
+            packetstream.setFloatingPointPrecision(QDataStream::DoublePrecision);
             double val;
             packetstream >> val;
-            if (qIsInf(val) || qIsNaN(val))
```
peterbarker: `qIsNaN` should return true for quiet NaNs.  You could avoid casting to `quint64` by doing the `qIsNaN` and doing the cast in the true case to see just what sort of NaN you're handing.

src/ui/Loghandling/BinLogParser.cpp
```diff
@@ -332,14 +349,32 @@ bool BinLogParser::parseDataByDescriptor(QList<NameValuePair> &NameValuePairList
...
+            quint64 *valPtr = reinterpret_cast<quint64*>(&val);
```
peterbarker: `const`?

src/ui/Loghandling/BinLogParser.cpp
Arne-W: Tried that on double and it did not work - but while reading your comment remembered that the double value was extracted wrong so the `qIsNaN` could not work.
So thanks for that hint - will change it!

src/ui/Loghandling/BinLogParser.cpp
Arne-W: :+1:

src/ui/Loghandling/BinLogParser.h
Arne-W: I think we should change it as soon as there is another autopilot :wink:
Thanks a lot for that review!

Arne-W:
Will create a new PR with those fixes tomorrow.

Arne-W:
@OXINARF @peterbarker It would be great if you could do a new review. I have changed the the implementation cause I found even more things which could be better by myself. Looks like its often better to do an own review after one night. I have also adapted some log messages and log levels.

Arne-W:
@OXINARF @peterbarker Thanks again for reviewing - and yes the casting works as expected :grinning: tested a few minutes ago with a log a user sent me.
But I will change the logging - soft NaNs are more often than expected and flooding the logs! Don't be afraid logging changes won't need a review I think :wink: 
@peterbarker I have seen some new message types like `MULT` `UNIT` and `FMTU` where can I find a description for them?

peterbarker:
On Tue, 23 Jan 2018, Arne Wischmann wrote:

> @peterbarker I have seen some new message types like MULT UNIT and FMTU
> where can I find a description for them?

I've just raised an issue for you.

PR #1090, opened 2018-01-25
Info View: fix set speed

AndKe:
Older set speed were updating an parameter. (deprecated method)
This change uses MAV_CMD_DO_CHANGE_SPEED


Arne-W:
@AndKe Are you sure this works with all vehicle types? Just asking cause the old code did it in different ways?

AndKe:
Tested with ArduCopter and ArduPlane (none of them worked before this patch.
In any case, changing parameters (a permanent change) from a part of GUI that issues instant commands is kind of dirty.

Arne-W:
All right - thank you for this fix :+1:

PR #1098, opened 2018-02-24
About Dialog: Fixed credits link

dcarpy:
Fixed link to credits, it is now
http://ardupilot.org/planner2/docs/credits-and-contributors.html.


Arne-W:
:+1:

Issue #1095, opened 2018-02-22
PX4FirmwareUploader.cc - Compile error


drobban:
**Background**
I'm working on an port of apm_planner to OpenBSD and any input you have about this is much appreciated as my ability to test this code is limited.
But I'm able to "test" by compiling in both Linux and OpenBSD.

**Environment**
Compiler: clang
OpenSSL: LibreSSL 2.6.3
OS: OpenBSD 6.2
apm_planner: latest commit https://github.com/ArduPilot/apm_planner/commit/8b66bf1d4dcc597937ccaae20421603f338a24fc

*Where*
https://github.com/ArduPilot/apm_planner/blob/master/src/ui/configuration/PX4FirmwareUploader.cc#L654

*Why*
missing declaration of EVP_PKEY_get0_RSA.

*Non confirmed solution*
There is however a EVP_PKEY_get1_RSA declared in <openssl/evp.h>
after changing that I was able to build a working version of apm_planner in OpenBSD 6.2

**Question**
What potential consequences does my fix have?
Do you have any other proposed solutions that can be applied in both OpenBSD, Linux & others?

Arne-W:
Looks like it is used to verify the downloaded FW image. Is it still possible to verify a valid FW with the new function?

This is what I found somewhere in the net but I don't know what they mean incrementing the ref count?
-----------------
Notice that the ref count is updated. There are then two getters:

RSA *EVP_PKEY_get0_RSA(EVP_PKEY *pkey)
RSA *EVP_PKEY_get1_RSA(EVP_PKEY *pkey)

Where EVP_PKEY_get1_RSA will call EVP_PKEY_get0_RSA and then increment the ref count. This is the only reason I can think of that these function have 1 and 0. 1 for functions that update the ref count and 0 for those that dont. "In accordance with the OpenSSL naming convention the key obtained from or assigned to the pkey using the 1 functions must be freed as well as pkey.
-----------------

drobban:
Thank you @Arne-W for the reply.

So if I understand this correctly:
In order to fully replace EVP_PKEY_get0_RSA with EVP_PKEY_get1_RSA
a equivalent solution would be to move the getter outside on previous row, get the key and after the verification on row 654 free it with a call to ` void RSA_free(RSA *rsa);`

something like this

`RSA * rsa_key = EVP_PKEY_get1_RSA(pkey);`

[...]

PR #1102, opened 2018-03-03
OpenBSD compatibility

drobban:
Update in readme on how to build apm_planner in OpenBSD 6.2

Solves issue #1095
Changed EVP_PKEY_get0 to get1 to make it usable with both libressl and openssl.
m_enable_version_check uninitialized evaluates true in OpenBSD.

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -13,6 +13,36 @@ LinuxBuild {
 	}
 }
 
+
+# BSD Build settings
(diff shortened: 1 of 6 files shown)


drobban:
Suggested future commits would be;
1. Fixing sound feature in OpenBSD.
2. Removing the PX4FirmwareUploader::checkCOA as reports indicates that this functionality is no longer used.

Issue #1103, opened 2018-03-03
Hide Donate Button


drobban:
When "Hide Donate Button" is checked nothing happens.

Pull request is sent.

PR #1105, opened 2018-03-03
Donate eval

drobban:
Solves issue #1103


Arne-W:
@drobban I have merged this by doing a cherry pick to change the commit message as we like to have it in a special format. I am really sorry that I somehow removed your name as author from the commit while doing this - it was not my intention :worried: I did this several times before and never lost the original author - weired - sorry for that.
Nevertheless thank you for that fix :+1:  and you can find the contribution guidelines here: [https://github.com/ArduPilot/apm_planner/blob/master/CONTRIBUTING.md](https://github.com/ArduPilot/apm_planner/blob/master/CONTRIBUTING.md) There should be a link to a table with the subsystem names.

PR #1112, opened 2018-03-23
Firmware: changed URL, .eu is discontinued

AndKe:
@peterbarker confirmed that this change is permanent. There may be a
redirect in the future, but it does not work yet.


Arne-W:
Thanks for this patch :+1:

billbonney:
@Arne-W It wasn't that is was discontinued was the issue,  but that was a bug introduced (and I missed when reviewing). It should always be ardupilot.org and not the redirected site address

PR #1113, opened 2018-03-24
Some fixes

Arne-W:
Several bug fixes - perhaps someone wants to review :smile:

### DIFF
--- a/src/logging.h
+++ b/src/logging.h
@@ -19,56 +19,43 @@
 
 #include <QDebug>
 #include <QtGlobal>
+#include <QLoggingCategory>
 
-// Defines for configuring the used loglevels
-// just uncomment if you want to hide one level
+// Declare the base logging category - creation is done in main.cc
+Q_DECLARE_LOGGING_CATEGORY(apmGeneral);
+
+// Define for enabling trace logging. As trace logging is not supported
+// by the Qt logging framework the trace level is handeled by the debug level.
+// Disabling the NO_TRACE define enabes the trace logging
 #define NO_TRACE
-//#define NO_DEBUG
-//#define NO_INFO
-//#define NO_WARN
-//#define NO_ERROR
 
-
-// logging macros
+// logging macro
 #ifdef NO_TRACE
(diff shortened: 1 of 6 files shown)


drobban:
commit https://github.com/ArduPilot/apm_planner/pull/1113/commits/bd5b7bd72e1ba1160480456fe7cde1b70ee16848 looks nice. Haven't tested it. Want to ask the question, why not use the built in functionality to suppress debug messages? Even if you want to keep the preprocess defines, why not use the `QT_NO_DEBUG` define instead of a new define as `NO_DEBUG`? What I can understand, you can still keep the conditions but instead of checking if `NO_DEBUG` has been defined, check if `QT_NO_DEBUG` is defined. This might confuse, as the `QT_NO_DEBUG` is defined in the release makefile but there is still debug messages comming out, otherwise one might instead ask the question, what does `QT_NO_DEBUG` do?
Might be wrong here and it wouldn't be the first time, but what are your thoughts about that?

Arne-W:
I do not really understand your question as the `NO_DEBUG` define is completely removed. The only "non Qt" define which is still there is the `NO_TRACE` as there is no corresponding log level in the Qt system. The other defines which can be used to turn of logging at compile time `QT_NO_DEBUG_OUTPUT, QT_NO_INFO_OUTPUT, QT_NO_WARNING_OUTPUT` are the official Qt defines. The `QT_NO_DEBUG` define changes the behaviour of the `Q_ASSERT()` macro but does not influence the logging output [see here for `QT_NO_DEBUG_OUTPUT`](http://doc.qt.io/qt-5/qtglobal.html#qDebug) and [here for `QT_NO_DEBUG`](http://doc.qt.io/qt-5/qtglobal.html#Q_ASSERT).
Does this clarify the change, or did I get you completely wrong?!

drobban:
No you are right @Arne-W must of been really tired when I checked that commit last night. sorry for that.
From my point of view, the change you made is an improvement. again sorry for the mix up.

Arne-W:
@drobban No problem - nice to see someone cares about and takes a look - and it is always good to discuss a solution :+1:

Issue #1089, opened 2018-01-23
Add support for using unit information from dataflash logs


peterbarker:
ArduPilot's dataflash logs now contain unit information for each and every field.

There are several new messages involved in this - but they act in the same way as `FMT` messages, being basic "header information".

The new `UNIT` message defines a mapping between a character and a unit.  These units are, for the most part, SI units - `volts`, `seconds`, `Hz` - and derived units such as `m/s`, however there are some "special case" units like `satellites` and `deglatitude`.  It is expected that a GCS be able to convert these into units presentable to a user (perhaps according to their preference), so for example temperature in Fahrenheit in Liberia.

The new `MULT` message defines a mapping mapping between a character and a multiplier.  So, for example, `2` indicates a multiplier of `100`.  These are used where the natural presentation of a time series in a log is not in the base SI unit - the obvious example being `TimeUS` - a time in microseconds since boot.

The new `FMTU` message provides the unit for each field present in the corresponding FMT message.  `FmtType` gives the `FMT` number, and the other two fields present both the unit and multiplier for the corresponding (by offset) field in the `FMT` message.  For example, almost every message we emit starts with a `TimeUS` field, a time since system boot in microseconds.  The `FMTU` message for each of these would have `s` in `UnitIds[0]` and `F` in `MultIds` (i.e. `10^-6`).  `F` is a reference into the table formed by `MULT` and `s` is a reference into the table formed by `UNIT`

```
2018-01-24 10:13:34.20: FMTU {TimeUS : 1687658, FmtType : 229, UnitIds : s--, MultIds : F--}
```

Possible uses include presenting fields in locale-specific forms, labelling graph axes, and possibly determining which time series might be presented on the same axis (given they have the same unit and a similar multiplier).  In tools where formulae can be used to present derived values it would be possible to enforce unit-correctness - only subtracting seconds from seconds and metres from metres, for example, and where the language permits adding units generally to values it would be possible to enforce comparisons only against like-units.

peterbarker:
@Arne-W Only the FMT units are guaranteed to be written at the moment.  Sorry.

Arne-W:
@peterbarker Ok - not a very big problem :wink: 
[...]

PR #1115, opened 2018-04-01
WIP: modify KML exporter to facilitate analysis of aerobatic maneuvers

kd0aij:
This PR automatically segments the flight path into "maneuvers" bounded by straight and level flight.
GE can animate this flightpath using a feature called "tours", and that might prove useful for log analysis.

Example KMZ file is here: https://drive.google.com/open?id=1DXqYigl2Eek_ZoIoXqK3eNfPJBOrbrEU

Download the file and open it in Google Earth, set the touring options as shown below:
touring_options

then select the "tour" named maneuvers (leave the checkboxes as they were) and click the "play tour" button at the bottom of the "Places" sidebar:
places_sidebar

Also, clicking on the flightpath will pop up info including RPY angles, AOA, and speeds (exported at 5Hz):
mdisplay

### DIFF
--- a/src/output/kmlcreator.cc
+++ b/src/output/kmlcreator.cc
@@ -153,6 +153,12 @@ Attitude Attitude::from(FormatLine &format, QString &line) {
     return a;
 }
 
+AoaSsa AoaSsa::from(FormatLine &format, QString &line) {
+    AoaSsa a;
+    a.readFields(format, line);
+    return a;
+}
+
 AHR2 AHR2::from(FormatLine &format, QString &line) {
     AHR2 a;
     a.readFields(format, line);
@@ -388,7 +394,7 @@ void KMLCreator::processLine(QString &line)
             }
         }
     }
-    // POS, ATT, AHR2, NKQ1, and XKQ1 messages are all logged at 25Hz (by default).
+    // POS, ATT, AHR2, AOA, NKQ1, and XKQ1 messages are all logged at 25Hz (by default).
     else if(line.indexOf("POS,") == 0 && (gpsOffset > 0)) {
         Placemark* pm = lastPlacemark();
         Attitude att;
@@ -424,7 +430,8 @@ void KMLCreator::processLine(QString &line)
 
             if(gps.hasData()) {
                 // add a pos/att pair to the ManeuverData object
-                m_maneuverData.add(gps, att);
+                m_maneuverData.add(gps, att, m_aoa);
+
                 m_summary->add(gps);
 
                 Placemark* pm = lastPlacemark();
@@ -481,6 +488,12 @@ void KMLCreator::processLine(QString &line)
             if(m_att.hasData()) {
                 m_newATT = true;
             }
+        }
+    }
+    else if((line.indexOf("AOA,") == 0)) {
+        FormatLine fl = m_formatLines.value("AOA");
+        if(fl.hasData()) {
+            m_aoa = AoaSsa::from(fl, line);
         }
     }
     else if(line.indexOf("CMD,") == 0) {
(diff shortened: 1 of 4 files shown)


kd0aij:
@lvale https://discuss.ardupilot.org/t/fun-with-df-logs-apm-planner2-kml-and-google-earth/27515

AndKe:
I like it. :)
It would make it even easier to understand some events.

AndKe:
here is what I see .. https://youtu.be/2n0eM7CG85E

kd0aij:
That's disappointing. Google Earth Pro doesn't even run properly on my Ubuntu 16.04 machine though. I used Google Earth 7.1.8.3036 (64-bit) with OpenGL. For some reason, you're seeing huge icons (they shouldn't overlap at 2 meter spacing). And the KML animation demos that I've tried recently don't work properly for me either, so I'm reluctant to invest a lot of effort in something that flaky.

Maybe it would be more worthwhile to switch to Cesium for this? The PX4 guys recently added Cesium animations to their log analyzer, and that is closer to what I wanted to achieve anyway: https://review.px4.io/3d?log=54f72ade-68f5-40b7-97e0-f51dcb95b7d2

Perhaps there is someone on the team who knows how to use Cesium?

AndKe:
@kd0aij  Yes, Also - I have both Google Earth 7.1.x and this Pro one I just installed ...
Also: the plane icon fly sideways...

I agree, using Cesium would solve a lot of this weird things, not to mention Google's sloppy Linux compatibility (I know others that had to hop thru a lot of rings to make Google Earth run on Linus)
I guess you wish to integrate it to https://github.com/goodrobots/MAVCesium ?   - or just use some Cesium and add it directly to apmplanner ? (that would be even more awesome.. )   :)

kd0aij:
@AndKe Thanks, I had forgotten that there was a mavproxy module for Cesium. That looks like a better way to go, and it might be easy to add an export from apmplanner.

Arne-W:
Cool - if you are convinced that Cesium is the better way to go we should do it. Adding a new log exporter should be easy (not taking into account the complexity of the exporter itself).
But for long term we should decide which format we will support and which format will not be maintained anymore. If we can support most of the KML features with cesium and add more cool features it would be perfect, but ending up in 2 export formats where none can substitute the other would be annoying.
Don't get me wrong - we do not have to make a perfect solution on first shot - it may take a while until we are able to remove, or at least stop maintainig one solution, but we should decide which way to go.

@billbonney As our chief mainatiner :wink: - what do you think?

@kd0aij If you want to start making a proof of concept be aware that the Logformat will support scaled data in the near future - don't know if it has an impact for such an exporter. If you need further information you should look here #1089. If you need further information let me know (implementation is nearly completed - exporter do not support it until now)

AndKe:
An idea;
What if AP2 did *not* export to mavcesium ?   - not export a file or a static pile of data.
What if; mavcesium got a "slave mode", where it just accepted a data stream from AP2 and just displayed it, regardless of timestamps or other checks.

The data from AP2 could then be sent to mavcesium as a "playback" that could be started, stopped, , reversed , moved step by step...  using controls in AP2

Think of it: one could "playback" the moment of a failure based on finding it in log, and step thru it, watching a graph cursor move as mavcesium would show it ...  every moment would be visible in graph with lots more related data from the same moment as cesium dislpayed it.. THAT would be totally awesome !  :)

Arne-W:
Nice Idea - BUT at the moment the log export works on dataflash logs not on Mavlink logs and there is no support in AP2 to convert dataflash to mavlink. Moreover it looks like a LOT of work and - to be honest - at the moment there are not enough developers supprting AP2 to implement such a cool big new feature....

But you are right - That would be awesome!

AndKe:
I did not mean to convert it to mavlink.. only send attitude, altitude,
position heading... And basic data for PFD .
And I do see it's a pile of work, but maybe less than inventing an export
plus import solution.

kd0aij:
@AndKe Lately I've been using GEpro on Win10 to analyze KML files generated in APM2, and the model scales are OK for me.  Maybe you're seeing a bug in the Linux implementation?
Here's an image capture from GEpro: 
outside_loop_apm2_kml
The KMZ file and image are available here:
https://drive.google.com/open?id=1Tf1z2o3WEpvyjXvtVmudd8SxZK3xxgzq

AndKe:
Yes, that is how it should work, the plane icon is HUGE on Linux version, I notified Google, but can't count on it being fixed anytime soon ... 
I assume the plane symbol is a native part of Google Earth ? - not something that can be changed ?

kd0aij:
No, the plane is a Collada model which is copied into the KMZ file from here:
https://github.com/ArduPilot/apm_planner/blob/262c6d2bdeb02d9ff9ad1d21260aff6eb64e3ccf/files/vehicles/block_plane/block_plane_0.dae
You could try changing the scaling factors in the KML, but I don't think that will fix the problem with GE.

lvale:
seems ok on MacOS GGL Earth Pro
screenshot 2018-10-31 at 13 59 38

PR #1118, opened 2018-04-15
Mission Widget: Terrain Alt will not loose markers anymore. Fixes #1117

### LINKED ISSUE #1117: waypoint markers gone if connected to vehicle.
In current master, once AP2 connects to a vehicle (sitl too) , the waypoints will loose the markers and  become un-dragable.
screenshot from 2018-04-13 07-03-57

Arne-W:
Bugfix PR.
@AndKe should fix the issue. BUT I do not know why this waypoint type was not handled - only forgotten or on purpose?! Don't if it only fixed the symptom or really the big - test carefully when using real UAV.


AndKe:
perfect !  
 thanks. :)

Issue #1121, opened 2018-04-27
Startup delay after use with Solo 


AndKe:
please see this delay while waiting for Solo...  (there is no trace of that in GUI/connections)

```
[20180427 23:04:19.508 DEBUG] - APMToolBar setConnection: false
[20180427 23:04:19.508 DEBUG] - APMToolBar: updateLinkDisplay "No Devices" : "115200"  connected: false
[20180427 23:04:19.509 INFO ] - UDP Created  "UDP Link (port:14550)"
[20180427 23:04:19.509 INFO ] - UDPLink::UDP connect  QHostAddress(QHostAddress::Any) : 14550
[20180427 23:04:19.510 INFO ] - UDP: ADDING HOST: "127.0.0.1:46650"
[20180427 23:04:19.510 DEBUG] - HOST:  "127.0.0.1"
[20180427 23:04:19.511 DEBUG] - Address: "127.0.0.1"
[20180427 23:04:19.522 INFO ] - UDP: ADDING HOST: "10.1.1.1:57060"
[20180427 23:04:19.522 DEBUG] - HOST:  "10.1.1.1"
[20180427 23:04:19.548 DEBUG] - TryCorrectGoogleVersions, VersionGoogleSatellite:  "798"
[20180427 23:04:24.527 DEBUG] - Address: "10.1.1.1"
[20180427 23:04:24.540 INFO ] - UDP: ADDING HOST: "10.1.1.1:57060"
[20180427 23:04:24.541 DEBUG] - HOST:  "10.1.1.1"
[20180427 23:04:29.545 DEBUG] - Address: "10.1.1.1"
[20180427 23:04:29.559 INFO ] - UDP: ADDING HOST: "10.1.1.1:57060"
[20180427 23:04:29.560 DEBUG] - HOST:  "10.1.1.1"
[20180427 23:04:34.562 DEBUG] - Address: "10.1.1.1"
[20180427 23:04:34.610 DEBUG] - SCW: Link is disconnected
[20180427 23:04:34.627 INFO ] - No Ports Found ()
[20180427 23:04:34.628 DEBUG] - Baud rate: 115200 Expected: 4


```

the .ini file remembers it all to well..
```

[LINKMANAGER]
LINKS\1\baud=115200
LINKS\1\linkid=1
LINKS\1\port=No Devices
LINKS\1\type=SERIAL_LINK
LINKS\2\HOSTS\1\host=127.0.0.1
[...]

PR #1126, opened 2018-05-17
Comms: linkmanager will only store one host IP for each UDP connection

Arne-W:
This is some sort of Hack. The Linkmanager and the UDPlink is able to handle more than on host IP making it possible to communicate with more than one UAVs simultaniously.
It seems that this feature leads to long startup times for some users see #1121.

I am not sure if this hack solves one issue by adding a new one for other users, as I do not use UDP links except for SITL!
Someone else should test and decide it this shall be merged!


drobban:
I use UDP link. Perhaps I can test some time during the week.

Arne-W:
@drobban no need to test this anymore. We will make a better solution for #1121

PR #1132, opened 2018-07-08
Mavlink2 migration

Arne-W:
Changed mavlink library to version 2.0 and adapted code base.
I hope I found all parts which had to be adapted.

If anybody wants to review just review the second commit as the first one is mainly the library replacement.


Arne-W:
cherry picked

PR #1133, opened 2018-07-16
Build: Added files generated during build to gitignore

hdino:
While building a freshly cloned repository I saw that some generated files are not tracked by the `.gitignore` file and thus clutter up the `git status` output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,12 +49,16 @@ user_config.pri
 *.suo
 *.uhf.txt
 *.opensdf
+*.qmlc
 apmplanner2.xcodeproj/
 GeneratedFiles/
 deploy/*.exe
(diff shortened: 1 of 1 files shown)


Arne-W:
On which plattform are you building? Cause I am building on linux (ubuntu 16.04) and my git status is clean after a build.

hdino:
I also build on Linux (Debian Buster). My qmake version is 5.10.1, maybe this causes the additional files being built.

Arne-W:
If you would do an "out of source" build by enabling "shadow build" and an additional build directory those files would be created in the build directory. You should try it as the compiler output does not pollute your source directory.
Nevertheless we can merge it making it easier to handle "in source" builds.

PR #1137, opened 2018-08-02
Mavlink 2.0 fixes

Arne-W:
WIP for making mavlink 1.0 communication work again


Arne-W:
added commit for better mavlink version detection. Version is now checked due to a Autopilot_Version message request and only switched to 2.0 if vehicle reports capability.

PR #1146, opened 2018-08-12
GraphView: Message IDs for Unit data are now setup dynamically.

Arne-W:
The old unit handling code had static defines for UNIT, MULT and FMTU messages, but the message ID of this messages are dynamic and can change.
Now the message IDs of this messages are set up while parsing.


Arne-W:
added 2 commits - one for detecting the STRT message by name and one to fix several compiler warnings.

Issue #1155, opened 2018-09-20
Graph: "Show Mode" does not show mode for loaded .tlog


AndKe:
it does, however, appear to work for graphing tlog data as they come in, but for logged graph

Arne-W:
Looking at the code I think it has never worked cause I used the wrong datafield for extracting the mode from the heartbeat message....
If you have some older logs where you can compare tlog and bin logs of the same flight it would be interesting if it has ever worked correctly?

Arne-W:
Thanks for testing - as you can see I made a fix.
Your last log should not have worked too - did the mode change and was it different from auto or guided?

AndKe:
now that you ask, I see there is something very strange with the "heli" finding.
It shows Guided and AltHold - but when I configure "table filter" for HEARTBEAT only, I can't find any of those changes in HEARTBEAT.custom_mode

Arne-W:
without the bugfix the mav type was interpreted as mode - you should see a change in data there.

AndKe:
yep, you are perfectly correct.  :)

Issue #1154, opened 2018-09-20
Graph: .tlog export to KML/KMZ


AndKe:
the export appears to work, but the resulting .kmz does not have any flight-based content (no positions/modes/etc)

Arne-W:
Exporting a tlog to a KML has never worked as the kml exporter tries to find special message names which are used in flashlogs. The messages in tlogs have different names...
The whole exporting for KML is not very good as we will get problems with scaled logs too. We need some sort of redesign...

Arne-W:
...perhaps we should disable the kml export for tlog files.

Arne-W:
Disabled the kml export for all but flashlogs. 
Fixed with e2a15de5294d8bd67be2cbc0d84606d663f20c47

PR #1156, opened 2018-09-21
Some Bugfixes

Arne-W:
First commit should fix issue #1155


Arne-W:
Added another commit fixing #1154 by disabling KML export for tlogs as it have never worked.
Adapted some doxygen commenting

PR #1161, opened 2018-12-05
MainWindow.cc: remove unused code in comments

bekicot:
While learning the code in MainWindow.cc. I saw a lot of commented code which may not be used. I also saw some of empty `if` that can possibly be removed. For me it is confusing.

In this pull request I did:
- Format comments (from `//comment` to `// comment`
- Remove commented code.

However, I'm coming from another language. My C knowledge is pretty much basic. If this is something expected, please close this pull request.

If you guys think this pull request helps. I will be happy help to clean up another comment, to make the code more readable. In my opinion, it will help separate between meaningful comment and some code that may or may not be used.

### DIFF
--- a/src/ui/MainWindow.cc
+++ b/src/ui/MainWindow.cc
@@ -292,7 +292,6 @@ MainWindow::MainWindow(QWidget *parent):
     widget->setMinimumWidth(1024);
     widget->setFeatures(QDockWidget::NoDockWidgetFeatures);
     widget->setTitleBarWidget(new QWidget(this)); // Disables the title bar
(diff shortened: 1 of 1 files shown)


billbonney:
there is some code in here that disables certain features that could be re-enabled. And that is commented out, but there is a comment to explain that in the code.

Arbitrarily changing the comment style just adds noise for already tested code. I think changes can happen as part of number of commits, but just whitespace is not going to be accepted.

But if you have improvement and changes we will happily accept them. Probably a good idea to raise an issue and we can discuss proposed changes you'd like to make

PR #1174, opened 2019-07-13
Copter: Change to setting parameter of AC 3.6.9 release version.

muramura:
1. I created the configuration parameter file for AC 3.6.9 release version with the following command.
./Tools/scripts/build_parameters.sh

2. I added the throwing mode and the zigzag mode to the flight mode selection item.


Arne-W:
Nice - thank you a lot! :+1:

PR #1175, opened 2019-07-15
Copter/CameraGimbal: Change from CH to RC, RC to SERVO

muramura:
1. I change from CH to RC. It became a radio control from the channel.
2. I change from RC to SERVO. It distinguishes RC and SERVO.
3. When SERVO is specified by the camera trigger of the camera gimbal, the setting item of the camera trigger can be changed.


Arne-W:
@muramura Thank you for participating to AP2. I will do the review as soon as possible.

muramura:
@Arne-W san... 
Thanks for your comment.
The following file is unnecessary.
I will delete it.
===

src/ui/configuration/CameraGimbalConfig.cc.bug

Arne-W:
Thank you 👍

PR #1177, opened 2019-07-28
map: Added Japan Map

muramura:
I live in Japan.
I have many mountainous areas in Japan.
I want to use a map with elevations.
I want to be able to use the map of the Geographical Survey Institute of Japan.

Screenshot from 2019-07-28 21-10-46


Arne-W:
thanks :+1:

PR #1179, opened 2019-08-23
read and request int32 waypoints

niuxw:
Hello APM Planner developers,

I found a problem that the waypoints drawn on the map are not accurate. According to this post: https://discuss.ardupilot.org/t/inaccurate-waypoint-logging-with-toggle-switch/26636, I believed that the error come from the conversion from float to double value. APM Planner requests and write waypoints gps position  in float, however, the Ardupilot system uses int32 to store the gps position for waypoints in default. So I changed the source code to make APM Planner request and write each waypoints in int32. The message ID for the mission item, which contains waypoint ID, should be MAVLINK_MSG_ID_MISSION_ITEM_INT rather than the MAVLINK_MSG_ID_MISSION_ITEM, and the mission item type should be mavlink_mission_item_int_t instead of mavlink_mission_item_t.

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -1052,14 +1052,26 @@ void UAS::receiveMessage(LinkInterface* link, mavlink_message_t message)
         }
             break;
 
-        case MAVLINK_MSG_ID_MISSION_ITEM:
+        // add MAV Message id: MAVLINK_MSG_ID_MISSION_ITEM_INT
+        // to handle the mavlink_mission_item_int_t type message
+        case MAVLINK_MSG_ID_MISSION_ITEM_INT:
         {
-            mavlink_mission_item_t wp;
-            mavlink_msg_mission_item_decode(&message, &wp);
-            //QLOG_DEBUG() << "got waypoint (" << wp.seq << ") from ID " << message.sysid << " x=" << wp.x << " y=" << wp.y << " z=" << wp.z;
+            mavlink_mission_item_int_t wp;
+            // using mavlink_msg_mission_item_int_decode to decode mavlink_mission_item_int_t type message
+            mavlink_msg_mission_item_int_decode(&message, &wp);
             waypointManager.handleWaypoint(message.sysid, message.compid, &wp);
         }
             break;
+
+        // unable to support float waypoints
+//        case MAVLINK_MSG_ID_MISSION_ITEM:
+//        {
+//            mavlink_mission_item_t wp;
+//            mavlink_msg_mission_item_decode(&message, &wp);
+//            //QLOG_DEBUG() << "got waypoint (" << wp.seq << ") from ID " << message.sysid << " x=" << wp.x << " y=" << wp.y << " z=" << wp.z;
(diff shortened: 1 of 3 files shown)


Arne-W:
Thank you for the change. Are you sure converting the waypoint from int to double for storing in APM Planner and later converting the back to int when writing them back to the UAS does not raise similar issues like the float conversion??

Arne-W:
@niuxw Hi - did you see my question?

hdino:
According to [https://en.wikipedia.org/wiki/Decimal_degrees](https://en.wikipedia.org/wiki/Decimal_degrees) or [this article](https://en.wikipedia.org/wiki/Wikipedia%3AWikiProject_Geographical_coordinates#Precision), 7 decimal places are sufficient for a minimum difference of ~11 mm at the equator.

A single precision float can store [6 to 9 significant decimal digits](https://en.wikipedia.org/wiki/Single-precision_floating-point_format#IEEE_754_single-precision_binary_floating-point_format:_binary32). So in the worst case, a float will yield only 11 cm precision.

APM Planner uses a double which can store 15 to 17 significant decimal digits.

hdino:
See also the green box under `MISSION_ITEM_INT vs MISSION_ITEM` in the [MAVLink docs](https://mavlink.io/en/services/mission.html#command_message_type).

Arne-W:
@hdino Did I get you right that you like this change? I am not really sure but if another developer says "yes it's good" I will merge it :smiley:

hdino:
@Arne-W I like the idea of supporting integer waypoints, because they offer higher accuracy. I'm just not sure if it would be better to implement a dual stack and let the user choose which type to use in the configuration. On the other hand, this would only make sense if there are autopilots that don't support integer waypoints.

Arne-W:
@hdino Ok - if I get you right the "perfect way" would be requesting and sending the waypoints as integer and if not supported by the autopilot as float while handling them internally as integer and only display them as double?

hdino:
@Arne-W Yes, the "perfect way" would be to transmit the waypoints as integer via MAVLink, probably with a fallback to float if integers are not supported by the autopilot. In APM Planner, however, they can be stored as double since double offers a higher accuracy than the int32 used in MAVLink.

Arne-W:
@hdino Thank you a lot for discussing this PR :+1: . I will merge it as improves accuracy and create a feature request for using float requesting as fallback.

niuxw:
Hi @hdino, thank you for your information. 7 decimal places have enough accurate, but the problem is when converting double to float (7 decimal digits of precision) in C++, the 7th decimal places may be changed sometimes. So it means FLOAT can only guarantee the accuracy of 6 decimal places.

niuxw:
Hi @hdino, I found an example for this conversion problem in C++. [http://www.cplusplus.com/forum/general/64482/](url)

niuxw:
Hi @Arne-W, I am so sorry for the late reply. From this website, [Converting from Double to Int](http://www.vias.org/cppcourse/chap03_02.html), conversion from int to double is safe, and conversion from double from int, the fractional part is discarded, which means the integer part won't be changed. So I think conversion int to double then back to int won't change int value.

PR #1181, opened 2019-09-16
Firmwareupdater: added support for flashing of CubeBlack.

AndKe:


Arne-W:
Cool thank you - sorry it took so long until I merged....

PR #1192, opened 2019-12-10
Vagrantfile: move to xenial64; write scripts for compilation

peterbarker:
The scripts make iterative development easier for those that don't
compile every day


Arne-W:
:+1:

PR #1194, opened 2020-01-09
mavlink: support both MISSION_REQUEST and MISSION_REQUEST_INT

hdino:
#1179 added support for integer waypoints. However, the code does not distinguish between a MAVLink MISSION_REQUEST and MISSION_REQUEST_INT.

I added this feature here. @Arne-W and @niuxw What do you think?

### DIFF
--- a/src/uas/UASWaypointManager.cc
+++ b/src/uas/UASWaypointManager.cc
@@ -1099,7 +1099,11 @@ void UASWaypointManager::sendWaypoint(quint16 seq)
             convertMavlinkMissionItem(wp, &wp_float);
             mavlink_msg_mission_item_encode(uas->getSystemId(), uas->getComponentId(), &message, &wp_float);
         }
-        else return;
+        else {
+            QLOG_DEBUG() << "sendWaypoint() - Current state does not allow sending waypoints. Check failed: current_state: " << current_state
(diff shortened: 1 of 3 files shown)


src/uas/UASWaypointManager.cc
```diff
@@ -1073,16 +1088,20 @@ void UASWaypointManager::sendWaypoint(quint16 seq)
...
-        emit updateStatusString(QString("Sending waypoint ID %1 of %2 total").arg(wp->seq).arg(current_count));
-
+        if (current_state == WP_SENDLIST_SENDWPSINT) {
+            mavlink_msg_mission_item_int_encode(uas->getSystemId(), uas->getComponentId(), &message, wp);
+        } else if (current_state == WP_SENDLIST_SENDWPSFLOAT) {
+            mavlink_mission_item_t wp_float;
+            convertMavlinkMissionItem(wp, &wp_float);
+            mavlink_msg_mission_item_encode(uas->getSystemId(), uas->getComponentId(), &message, &wp_float);
+        }
+        else return;
```
Arne-W: Perhaps we should print a logmessage (debug level) to show that something went wrong, instead of simply do nothing??

hdino: Indeed, that's a good idea :)

Arne-W:
Thanks for the support :+1:

PR #1201, opened 2020-04-03
Radio-Settings: added support for newer SiK versions

AndKe:
Fixed telemetry radio configuration so it works with SiK 2.0 and it's derivatives.
Since 2.0, noone could configure or see radio settings in GUI.
This solution is tested with 1.9 (same format as previous) , 2.0, and 2.0variants.

Please make a new release.

### DIFF
--- a/src/ui/configuration/Radio3DRSettings.cc
+++ b/src/ui/configuration/Radio3DRSettings.cc
@@ -118,25 +118,21 @@ bool Radio3DREeprom::setParameter(QString &parameterString)
 
 bool Radio3DREeprom::setVersion(const QString &versionString)
 {
-    QLOG_DEBUG() << "Radio3DREeprom Ver:" << versionString;
-    QStringList values = versionString.split(" ", QString::SkipEmptyParts);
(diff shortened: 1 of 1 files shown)


src/ui/configuration/Radio3DRSettings.cc
```diff
@@ -738,7 +734,7 @@ void Radio3DRSettings::writeEscapeSeqeunceNow()
 void Radio3DRSettings::readLocalVersionString()
 {
     QLOG_DEBUG() << "Radio3DRSettings::readLocalVersionString()";
-    emit updateLocalStatus(tr("Read local radio version"), black);
+    emit updateLocalStatus(tr("Read local radio versionDDDDDDDDDDD"), black);
```
Arne-W, line 737: Are those DDDD intended?

AndKe: Must be a new coding standard or something... I'll try to fix it ASAP

AndKe: It should be ok now.

PR #1202, opened 2020-04-03
Telemetry radio flashing fixed, and updated to SiK v2.0

AndKe:
Updated to 2.0 firmware, fixed flashing to work again.


src/ui/configuration/RadioFlashWizard.cpp
```diff
@@ -52,9 +52,9 @@ RadioFlashWizard::RadioFlashWizard(QWidget *parent) :
     ui->baudrateComboBox->addItem(QLatin1String("38400"), QSerialPort::Baud38400);
     ui->baudrateComboBox->addItem(QLatin1String("19200"), QSerialPort::Baud19200);
     ui->baudrateComboBox->addItem(QLatin1String("9600"), QSerialPort::Baud9600);
-    ui->baudrateComboBox->addItem(QLatin1String("9600"), QSerialPort::Baud4800);
-    ui->baudrateComboBox->addItem(QLatin1String("9600"), QSerialPort::Baud2400);
-    ui->baudrateComboBox->addItem(QLatin1String("9600"), QSerialPort::Baud1200);
+    ui->baudrateComboBox->addItem(QLatin1String("4800"), QSerialPort::Baud4800);
+    ui->baudrateComboBox->addItem(QLatin1String("2400"), QSerialPort::Baud2400);
+    ui->baudrateComboBox->addItem(QLatin1String("1200"), QSerialPort::Baud1200);
```
Arne-W, line 57: Nice finding

AndKe: Yes, I missed only huge DDDD's :)  
The explanation for those D's is that there were two of the same status/debug messages. As a newbie to QT, the simplest way was to change one text in an obvious way, (so I would not miss it later)  fantastic plan .... shameful execution.   :)

PR #1205, opened 2020-05-14
Vagrantfile: move to Bionic

peterbarker:
Focal doesn't run on my machine as a VM.  So move forward to something I've
tested....


Arne-W:
👍

PR #1206, opened 2020-05-14
ui: Radio3DRSettings: don't use qAsConst

peterbarker:
This doesn't work trivially on either of Xenial64 or Bionic64 installs


Arne-W:
:+1:

PR #1207, opened 2020-05-14
Added mission type to MAVLink mission_count message

hdino:
This pull request fixes the following bug:

While uploading the waypoints to the UAS, a `mission_count` message is sent. The struct `mavlink_mission_count_t` has a field `mission_type`, which was not set and therefore had a random value (so in most cases 0). You can find the possible values in [MAVLink's common.h](https://github.com/ArduPilot/apm_planner/blob/master/libs/mavlink/include/mavlink/v2.0/common/common.h#L664-L666). I set it to `MAV_MISSION_TYPE_MISSION`.

Best,
Dino

### DIFF
--- a/src/uas/UASWaypointManager.cc
+++ b/src/uas/UASWaypointManager.cc
@@ -1034,6 +1034,7 @@ void UASWaypointManager::sendWaypointCount()
     wpc.target_system = uasid;
     wpc.target_component = MAV_COMP_ID_MISSIONPLANNER;
     wpc.count = current_count;
+    wpc.mission_type = MAV_MISSION_TYPE_MISSION;
 
     emit updateStatusString(QString("Starting to transmit waypoints..."));
 


hdino:
@peterbarker Indeed! I guess the best way would be if the MAVLink C++ types had a constructor that takes values for all message fields as parameters. Then it's impossible to miss one (as long as the default constructor isn't used...).

peterbarker:
On Thu, 14 May 2020, Dino Hüllmann wrote:

> @peterbarker Indeed! I guess the best way would be if the MAVLink C++ types had a constructor that takes values for all message fields as parameters. Then
> it's impossible to miss one (as long as the default constructor isn't used...).

Very few projects use the C++ bindings.  I haven't checked - but these 
look like the C bindings to me.

Peter

hdino:
Yes, the C++ bindings are basically the same as the C bindings. I use them in my projects because then the MAVLink stuff doesn't flood the global namespace. However, they don't provide constructors other than the default one. So I suggest asking the pymavlink developers what they think about adding an additional value constructor.

Currently, some MAVLink types are initialized with memset(0), but 0 is probably not the value you want.

Do you have another idea, how the messages could be reliably initialised?

Arne-W:
I just can repeat peter - good finding! Thanks for that!

PR #1208, opened 2020-05-14
Allow MAV_CMD with values > 255

hdino:
The new MAVLink version uses an [uint16 to store commands](https://github.com/ArduPilot/apm_planner/blob/master/libs/mavlink/include/mavlink/v2.0/common/mavlink_msg_mission_item.h#L16) (`MAV_CMD`). However, in the current master the value for commands is [limited to 255](https://github.com/ArduPilot/apm_planner/blob/master/src/ui/mission/QGCMissionOther.ui#L53). This pull request sets it to [the maximum allowable value](https://github.com/ArduPilot/apm_planner/blob/master/libs/mavlink/include/mavlink/v2.0/ardupilotmega/ardupilotmega.h#L215).


Arne-W:
Nice - Thank you!

PR #1218, opened 2020-08-11
Firmware Install : Work around non-instant modem availability for PX4

### LINKED ISSUE #731: Pixhawk firmware update does not work on Ubuntu
Generally, connecting to Pixhawk is only possible about 10-15 seconds after connecting USB on Ubuntu which also leads to FW updates failing. I suspect, the UDEV daemon grabs the device first and only after it cannot find a suitable driver, it releases the device.
Strictly this is probably not an AP2 bug, but there should either be a solution in the documentation or maybe a suitable blacklist file in the Linux package.
I will try to investigate what exactly causes the problem and post a solution as I find it.

mbuzdalov:
During firmware updates, one is typically requested to replug the device.

On my system, the newly inserted device is nearly immediately detected by the planner, but the permissions on that device are not immediately set the right way (probably because `mtp-probe` needs to check it first). For this reason, firmware flashing is basically impossible.

This pull request introduces waiting for the device to become available with one-second sleeps between the attempts. While this is a dirty-ish hack, and something should have maybe done project-wide instead, I am not confident enough to do that. The introduced change just works for my use case, and is extremely unlikely to break anything.

This may probably fix #731, if I have guessed right which subsystem is responsible for that issue.

### DIFF
--- a/src/ui/configuration/PX4FirmwareUploader.cc
+++ b/src/ui/configuration/PX4FirmwareUploader.cc
@@ -289,9 +289,17 @@ void PX4FirmwareUploader::kickOffTriggered()
     m_port->setPortName(m_portToUse);
 #endif
 
-    if (!m_port->open(QIODevice::ReadWrite))
-    {
-        QLOG_ERROR() << "Unable to open port:" << m_port->errorString();
+    for (int t = 0; t < 100; ++t) {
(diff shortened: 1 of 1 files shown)


Arne-W:
After a long time - nevertheless thanks a lot!

PR #1229, opened 2021-01-31
General: Allow changing the Mavlink ID of APM Planner in advanced config dialog. Fixes #1228

### LINKED ISSUE #1228: Two AP2 GCS causes trouble for each other.
This is verified with current master. 2.0.28-rc1
Please see attached tlogs:
[2021-01-28 08-.zip](https://github.com/ArduPilot/apm_planner/files/5885594/2021-01-28.08-.zip)

The first tlog, (08:39) is created when the first AP2 GCS is started - referred to as "GCS1"
GCS1 loads all parameters just fine.
The second tlog, (08:40) is created when the first AP2 GCS is started - referred to as "GCS2"
GCS2 connects and start to load parameters.
GCS1 instantly starts to animate its parameter bar, with a count that is >100 behind. (so it does reload, but not by reading the same requests/responses)
GCS1 mavlink inspector discovers another MAVID 252 (GCS2?)
GCS2 mavlink inspector discovers another MAVID 252 (?)
none manage to fully load parameters. (give up/time out without message)
I end the test.

Arne-W:
Mavlink ID of APM Planner is configurable now.

### DIFF
--- a/src/QGC.h
+++ b/src/QGC.h
@@ -61,7 +61,7 @@
 
 namespace QGC
 {
-const static int defaultSystemId = 252; // Using 252 to 'crudely' identify a log created by APM Planner 2
+
 const static int defaultComponentId = MAV_COMP_ID_PRIMARY; // The main component ID is 1 for autopilot/and GCS (0 means all components)
 
 const QColor colorCyan(55, 154, 195);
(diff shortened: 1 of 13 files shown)


drobban:
Looks good, got worried when I first noticed the removal of 
m_enable_version_check(false) on line 45 inside src/comm/MAVLinkProtocol.cc
But then discovered that you initialized all variables in src/comm/MAVLinkProtocol.h.
Is this the preferred way of initializing variables in the future?

Arne-W:
@drobban - Yes since c++ 11 all primitives can be initialized in the header file. It works like using the initializer list but must not be done in every constructor - if you have more than one.

PR #1241, opened 2021-05-16
Allow 2000m in PFD altitude indicator. Fixes #1239

### LINKED ISSUE #1239: PFD Altitude >1000m - display error.
Screenshot from 2021-04-17 19-45-43

The altitude tape ends, and both the first digit and the decimal is gone.
A quick fix could be to align the text so that the four first digits were visible (and the decimal gone)

Arne-W:
Like the title says...


Arne-W:
@billbonney Could you please have a look - those qml things are somehow mysterious to me.

PR #1254, opened 2022-02-22
Feature/instance handling

Arne-W:
Added support for indexed data types in flash logs


Arne-W:
void DataSelectionScreen::enableItem(QString name) and
void DataSelectionScreen::disableItem(QString name) are failing when index is used!

Arne-W:
everything should work now...

PR #1266, opened 2022-10-25
Update mavlink version

andyp1per:


libs/mavlink/include/mavlink/v2.0/ardupilotmega/mavlink.h
```diff
@@ -28,19 +28,7 @@
...
-#endif
-
 #include "version.h"
 #include "ardupilotmega.h"
 
-#if defined(__GNUC__) && (__GNUC__ > 8)
-    #pragma GCC diagnostic pop
-#endif
-
-
```
Arne-W, line 45: Could you please migrate this change into the new version?? Without it the build generates thousands of warnings with newer gcc compilers. Moreover it would be very cool if you could add

`#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"`
`#pragma GCC diagnostic ignored "-Wuninitialized"`

to the first block to silence all warnings emitted by mavlink.

Arne-W (changes requested):

libs/mavlink/include/mavlink/v2.0/ardupilotmega/mavlink.h
andyp1per: @peterbarker this should be part of the mavlink generation step in AP - any idea how we can fix that up?

andyp1per: @Arne-W we think that this problem was largely resolved in mavlink at the end of last year, did you actually try the update?

Arne-W: @andyp1per Yes I built the change and got round about 3000 warnings. But to be absolutely sure I will check it again and inform you about the compiler and the options we are using.

andyp1per: Ok, thanks - I will double check the versions of the generator I used just to make sure I didn't miss a version

Arne-W: So I built from scratch on a new Kubuntu 22.04 with gcc 11.3.0 and I get more than 4000 warnings (counting all the duplicates).
```
mavlink_msg_obstacle_distance.h:173:158: warning: taking address of packed member of ‘__mavlink_obstacle_distance_t’ may result in an unaligned pointer value [-Waddress-of-packed-member]
```
We are building with -Wall -Wextra. Adding the diagnostic ignore pragma removes all warnings generated by mavlink.

Arne-W (approved):

PR #1267, opened 2022-10-26
Vagrant: use focal, Qt deps to 5.15.2, use install-qt

peterbarker:


Arne-W:
Nice - thank you for the update!

PR #1297, opened 2023-08-29
Mac refresh the mac build to deploy app and dmg

Iragne:
update the repo with the needed information related to the creation of an app Signed to be deployed on mac 
still under draft for now


apm_planner.pro
```diff
@@ -177,7 +177,7 @@ MacBuild {
     QT += multimedia
 
     QMAKE_INFO_PLIST = Custom-Info.plist
-    CONFIG += x86_64
+    CONFIG += x86_64 arm64
     CONFIG -= x86
     QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.7
```
Arne-W, line 182: I am not sure but I think its worth a test. Please remove the 2 CONFIG lines and use ``QMAKE_APPLE_DEVICE_ARCHS = x86_64 arm64`` instead and use ``QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.14``  I think mojave is old enough.
Perhaps this allows building a Universal binary

Iragne: Yes i was on it but look like i should to that 
CONFIG -= x86
CONFIG -= x86_64
CONFIG += x86 arm64

Iragne: Let me find an associated mac, I will also try to create different build and we can all test

Youw: Increasing the `MACOSX_DEPLOYMENT_TARGET` doesn't automatically enables Universal binary
that is a separate option
more than that, when building an arm64 or a universal binary - the niminum supported/deployement target is maxed to at least 10.14 automatically an w/o a warning, but only the the arm64 part of the binrary

PR #1299, opened 2023-09-10
Fix compilation with newer MSVC

Youw:
- there is nothing specific to MSVC2013 in the code, the project compiles/works with MSVC2019 just fine;
- MSVC-version specific deployement of runtime for version 2015 and higher is not implemented - to be done separately if needed;

---
This is not to claim the support of newer MSVC, but rather to allow local development/changes/experimantation.


Youw:
NOTE: tested with MSVC2019 + Qt 5.15.2.

Arne-W:
Nice - thank you for the contribution. Good to know the project can still be build on Windows!

Issue #928, opened 2016-04-13
Enhancement Request to Support HDPI Monitors


dcarpy:
Not sure how much work this is, but I thought I would put in an enhancement request.  Please see http://doc.qt.io/qt-5/highdpi.html for more info.

PR #1300, opened 2023-09-10
Support for HighDPI screens/monitors

Youw:
Relates to: #928

---
Essentially, this does not brings a full support of HighDPI screens.
There're still some visual artifacts, like low-resolution resources/images scaled up. Eacho one of those needs to be reviewed/handles separately.
But for the most part, because the original Application is written very well, all of the UI (as far as I could check myself) is scaled much better, so that all elements are of more comfortable size.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -110,6 +110,8 @@ int main(int argc, char *argv[])
     QCoreApplication::setAttribute(Qt::AA_X11InitThreads);
 #endif
 
+    QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+
     // Init application
     QGCCore core(argc, argv);
 


Youw:
To test this locally - one may simply set the `QT_ENABLE_HIGHDPI_SCALING=1` environment variable even w/o this change with Qt 5.14 or newer (the code change would work with Qt 5.6 or newer).

Arne-W:
@Youw  I have read about this parameter and tried it but for me it makes no difference on my 4k monitor. So where do I have to look ?

Youw:
what Qt version did you build your app with? Qt documentation claims if would work only with Qt 5.14 or newer

With my local build (and w/o the update suggested in the PR) with Qt 5.15.2 what I do is (from terminal/cmd on Windows):
```cmd
set QT_ENABLE_HIGHDPI_SCALING=1
apmplanner2.exe
```

Arne-W:
I am using Qt 5.15.3 on Ubuntu 22.04. Perhaps it is OS related that does not make a difference for me?

Youw:
Yes, it might be.
I don't have a Linux machine at the moment to play with it. Does the patch suggested in this PR changes anything for you on Linux?

You might as well try setting a different environment variable: `export QT_SCALE_FACTOR=2`
This is not recommended as a "production" solution, so I'm not suggestion it as a PR (the actual scale factor might be calculated in runtime), but rather a workaround to be used by "advanced" users locally on their PCs.

PR #1304, opened 2023-10-22
ui: Addressing fall through warnings

muramura:
There was a fall through warning during build.
I will change this warning.


Arne-W:
@muramura Hi, thank you for your effort - I will merge your change as having less warning is always good. Nevertheless qcustomplot is external code and therefore we normally do not fix warnings within it.

PR #1318, opened 2025-07-05
Fixes #797: Update UI layout for onboard compass calibration

### LINKED ISSUE #797: support for on-board compass calibrator
We have the new on-board compass calibrator in master now and it will also be in the 3.4.0 plane release. You can see the protocol here:
https://github.com/diydrones/ardupilot/blob/master/libraries/GCS_MAVLink/message_definitions/ardupilotmega.xml#L713
and an implementation in MAVProxy here:
https://github.com/DroneCode/MAVProxy/blob/master/MAVProxy/modules/mavproxy_calibration.py#L115
the on-board mag cal can do soft-iron corrections which can be very worthwhile on some aircraft.
It would be great to get support in AP2 for this new calibrator.

billbonney:
Added UI to allow for onboard compass calibration.

TODO: update plot to show report information during progress.

### DIFF
--- a/src/ui/configuration/CompassConfig.cc
+++ b/src/ui/configuration/CompassConfig.cc
@@ -23,8 +23,9 @@ This file is part of the APM_PLANNER project
 #include "logging.h"
 #include "CompassConfig.h"
 #include "CompassMotorCalibrationDialog.h"
-#include <qmath.h>
+#include "CompassOnboardCalibrationDialog.h"
 #include "QGCCore.h"
+#include <qmath.h>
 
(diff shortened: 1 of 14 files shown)


src/ui/configuration/CompassConfig.cc
```diff
@@ -350,6 +355,12 @@ void CompassConfig::liveCalibrationClicked()
     QTimer::singleShot(1000,this,SLOT(startDataCollection()));
 }
 
+
+void CompassConfig::onboardCalibrationClicked()
```
billbonney, line 359: delete method. unused.

